PR #38, opened 2015-04-05
Add meek fronted requests test specification

anadahz:


test-specs/ts-013-meek_fronted_requests.md
```diff
@@ -0,0 +1,108 @@
...
+	DomainName:HostHeader
+
+Example:
+
+	www.google.com:meek-reflect.appspot.com
+	a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net
+
+# Test description
+
+Performs a HTTP GET request to a list of fronted domains with the Host
```
hellais: It should be specified that the request is in fact done over a TLS encrypted session.

test-specs/ts-013-meek_fronted_requests.md
```diff
@@ -0,0 +1,108 @@
...
+## Required output data
+
+* The domain name and host header used in the measurement
+	(DomainName:HostHeader)
+
+* The requests that have been made
+
+* The received responses
+
+* If the meek request is blocked
```
hellais: This should probably read:

"If the meek **server** is blocked"

test-specs/ts-013-meek_fronted_requests.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Semantics
+
+censored:
```
hellais: Usually for boolean flags that indicate censorship we use the more neutral phrase "success", that can be True or False.

This is because censorship is a matter of intent and we don't know if what is happening is in fact censorship or just a misconfiguration of some network device.

test-specs/ts-013-meek_fronted_requests.md
```diff
@@ -0,0 +1,108 @@
...
+Example:
+
+	www.google.com:meek-reflect.appspot.com
+	a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net
+
+# Test description
+
+Performs a HTTP GET request to a list of fronted domains with the Host
+Header of the "inside" meek-server. The meek-server handles a GET request and
+response with: "I’m just a happy little web server.\n".
```
hellais: This needs to include more details as to what exactly is done by the test when doing the request. This includes eventual HTTP headers that are set by the client, what TLS parameters are used for the handshake.

hellais:
Ok I will merge this and your fix for it.

PR #41, opened 2015-04-14
Feature/test info

hellais:


hellais:
I think if anything this should live inside of `ooni-web`. Closing this as superseeded by: https://github.com/TheTorProject/ooni-web/issues/235.

PR #43, opened 2015-11-01
 Rename test spec to conform with name test spec

anadahz:


hellais:
lgtm. Thanks a bunch :smile:

PR #44, opened 2015-11-11
add psiphon test specification

juga0:
* add initial version of add psiphon test specification
* add ConnectionDone error in df-000-base.md. This error is triggered when the URL in a HTTP request is not found using psiphon


hellais:
Great work. Thanks a lot!

PR #47, opened 2015-12-21
Fix/lantern test spec

anadahz:


anadahz:
Rebased to current master. @hellais

PR #48, opened 2016-01-25
Add draft of web_connectivity test

hellais:


test-specs/ts-016-web-connectivity.md
```diff
@@ -0,0 +1,175 @@
...
+
+A list of URLs to be tested.
+
+## Semantics
+
+One URL per line:
+
+http://example-1.com/
+http://example-2.com/path1/
+http://example-3.com/path2/
```
bassosimone: Is the test expected to also test `https` URLs?

test-specs/ts-016-web-connectivity.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Test description
+
+This test is divided into multiple steps that will each test a different aspect
+related to connectivity of the website in question.
+
+The first step is to inform the web_connectivity test helper of our intention
+to the measurement against the URL in question and hence have it perform a
```
bassosimone: To "do" or "perform the measurement", I guess

test-specs/ts-016-web-connectivity.md
bassosimone: Responding to myself, yes. Maybe it could be interesting to add a step that establishes an SSL session and gathers information about the certificate being used.

test-specs/ts-016-web-connectivity.md
```diff
@@ -0,0 +1,175 @@
...
+   Record the list of A records in a list inside the report under the key "ips".
+
+3. **TCP connect**
+   Attempt to establish a TCP session on port 80 for the list of IPs identified
+   at step 2. If the URL begins with HTTPS then also establish a TCP session
+   towards port 443.
+
+4. **HTTP GET request**
+   Perform a HTTP GET request for the path specified in the URI and record the
+   response.
```
bassosimone: With @hellais we discussed whether to keep the mechanism used by http-requests to randomize the host header or not, and we concluded it was better to stick to the most common host header and always use that, to avoid the issue that, depending on the host header (e.g. mobile or not), the site may respond differently.

bassosimone:
Ah, and overall I really like this new test! :+1:

test-specs/ts-016-web-connectivity.md
anadahz: Since the HTTP requests test has support for HTTPS URLs it should be mentioned mentioned in the examples of the semantics section.
Indeed adding support for TLS/SSL session info will be very valuable addition to the web connectivity test; perhaps we should add the [tls_handskake](https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettests/experimental/tls_handshake.py) test?

PR #49, opened 2016-02-01
Feature/df 0.2

hellais:
In here we update the dataformat to reflect the normalisations done by the pipeline as well as the changes done to ooni-probe.


hellais:
Since the JSON reports branch has been merged we can proceed to merging this as well.

PR #50, opened 2016-03-15
Feature/fix cct tests

hellais:
This goes together with https://github.com/TheTorProject/ooni-probe/pull/463


anadahz:
Thanks for getting the specs pages updated. These changes however should refer any changes to the user documentation.
In the future you should do these as separate PRs as it makes reviewing easier.

PR #55, opened 2016-05-02
Update legal section of informed consent document

agrabeli:
I've updated the legal section of OONI's informed consent policy.


agrabeli:
This information exists in risks.md.

PR #61, opened 2016-05-26
Wrote a description for the web-connectivity test

agrabeli:


test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,97 @@
+# Web connectivity
+
+This test examines whether websites are reachable and if they are not, it
+attempts to determine whether access to them is blocked through DNS tampering,
+TCP connection RST/IP blocking or by having a transparent HTTP proxy.
```
bassosimone: Here I'd say just "or by a transparent HTTP proxy"

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,97 @@
...
+
+* Resolver identification
+
+* DNS lookup
+
+* TCP connect
+
+* HTTP GET request
+
+By default, this test performs the above both over a control server and over the
```
bassosimone: I am not sure "the test performs the above" is totally accurate here: resolver identification should only performed over the network of the user. Perhaps you can give numbers to steps and say that steps 2-4 are performed over both networks?

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,97 @@
...
+clear sign of network interference; but if the results are different, then the
+websites that the user is testing are likely censored.
+
+Below we provide information about how each step performed under the web
+connectivity test works.
+
+## 1. Resolver identification
+
+The domain name system (DNS) is what is responsible for transforming a host name
+(e.g. torproject.org) into an IP address (e.g. 38.229.72.16). ISPs, amongst
```
bassosimone: Perhaps we could expand ISP here because it's the first time it appears?

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,97 @@
...
+others, run DNS resolvers which map IP addresses to host names. In some
+circumstances though, ISPs map the wrong IP addresses to the wrong host names,
```
bassosimone: Uhm, don't they map the requested host name to the wrong IP?

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,97 @@
...
+attempting to establish a TCP session on port 80 (or port 443 for URLs that
+begin with HTTPS) for the list of IP addresses that were identified in the
+previous step (DNS lookup).
+
+## 4. HTTP GET request
+
+As the web connectivity test connects to tested websites (through the previous
+step), it sends requests through the HTTP protocol to the servers which are
+hosting those websites. A server normally responds to an HTTP GET request with
+the content of the website that it is hosting.
```
bassosimone: I'd say "with the content of the webpage that is requested"

bassosimone:
Apart from minor comments, this is very good!

In my opinion it could be improved by providing an example. I am not sure whether it would be better with introducing the example before discussing each step and then, for each, step discussing what the step practically means in terms of the example, or whether it would be better to have just an example section. In both cases, I think it would be interesting if what is printed by ooni-probe when running the test matches what we say here (which is quite already the case) such that nontechnical users could see 1:1 mapping between what the documentation says and what the program says. Does this make sense?

test-descriptions/web-connectivity.md
agrabeli: Edited.

test-descriptions/web-connectivity.md
agrabeli: Edited.

test-descriptions/web-connectivity.md
agrabeli: Edited.

test-descriptions/web-connectivity.md
agrabeli: Yep! Edited.

test-descriptions/web-connectivity.md
agrabeli: Edited.

agrabeli:
Yep, totally makes sense, and that's a great suggestion! Just added some examples.

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,148 @@
...
+    if one or more of the following applies:
+
+    * The body length of compared websites (over the control server and the
+      network of the user) differs by some percentage
+
+    * The HTTP headers do not match
+
+    * The HTTP status codes do not match
+
+    * The HTML title tags do not match
```
hellais: Actually the status code MUST always match and one between the HTTP headers or the HTTP body length or the HTML title tag must match.

hellais:
LGTM!

test-descriptions/web-connectivity.md
```diff
@@ -0,0 +1,146 @@
...
+
+    * The body length of compared websites (over the control server and the
+      network of the user) differs by some percentage
+
+    * The HTTP headers names do not match
+
+    * The HTML title tags do not match
+
+The examples below (testing piratebay.se and google.com for censorship in Italy) show
+what the output of the web connectivity test could look like:
```
bassosimone: Awesome!

PR #62, opened 2016-05-27
Wrote a consent form which can be used in the GUI and published on th…

agrabeli:
…e main website

### DIFF
--- a/informed-consent/consent-form.md
+++ b/informed-consent/consent-form.md
@@ -6,11 +6,13 @@ Project](https://www.torproject.org/), which collects and processes network
 measurements with the aim of detecting network anomalies, such as censorship and
 traffic manipulation. 
 
-Running OONI's software tests can potentially be a *risky* activity, depending on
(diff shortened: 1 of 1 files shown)


anadahz:
There is already an [informed consent](https://github.com/TheTorProject/ooni-spec/tree/master/informed-consent). I think it makes sense to edit the specific files and maybe add new as needed in the 'informed-consent' directory.

hellais:
@anadahz this is something different though. It's the actual form that will be presented to the user upon first running ooniprobe (or on installation) to agree express that they understand the risks associated with performing network measurements with ooniprobe.

The other documents inside of informed consent are too verbose to be shown to every user and we should probably be cross referencing them in the form for people that want more information.

The `informed-consent-long`, I believe, should be entirely removed in favour of this.

anadahz:
@hellais that makes more sense we should update/remove/merge what is needed not needed and place these files in the relevant directory.

hellais:
This is now inside of the ooni web UI. I think we can close this PR.

PR #68, opened 2016-06-04
oonib.md: remove a lie

bassosimone:
It is actually the converse. We may want to totally get rid of YAML.


hellais:
Indeed. We ended up going for the opposite approach. The reasoning there was that if we were to stick with YAML a the default data exchange format it added un-necessary complexity to serialise the YAML data and put inside of JSON.

oonib.md
```diff
@@ -61,10 +61,6 @@ using Tor or obfsproxy.
 Unless otherwise stated all of the network operations below can be performed
```
IstiaqSinn: I'm obvious

PR #70, opened 2016-06-12
Broken link in `inform-users-long.md`

nogoegst:
It probably should point to `/data` now.


hellais:
Thanks for spotting this.

Merging.

PR #72, opened 2016-08-02
Gui/quiz questions

agrabeli:
These quiz questions could be included in the 'informed consent' wizard that appears right before users can get started with the GUI. Users would have to get these questions right as a prerequisite to installing and running ooniprobe. 

First the wizard would include the 'informed consent' documentation that users would have to read and consent to. Then, once they're provided their consent, they would have to answer the 5 quiz questions correctly. Once they've done this, they can then start installing ooniprobe and running tests of their choice via the GUI. 

These quiz questions were developed with the aim of addressing the following:
1. Checking whether users understand what OONI software tests are designed to do
2. Checking whether users understand associated potential risks
3. Checking whether users understand that by running OONI tests via the GUI, their measurements will by default get published (which might also potentially include risks)

All feedback would really be appreciated! :)

Thanks.

### DIFF
--- a/informed-consent/quiz.md
+++ b/informed-consent/quiz.md
@@ -5,58 +5,58 @@ ooniprobe!
 
 ## Which of the following statements is true?
 
-[] OONI software tests are designed to detect censorship and surveillance
+[ ] OONI software tests are designed to detect censorship and surveillance
 
-[] OONI software tests are designed to detect censorship and signs of
+[ ] OONI software tests are designed to detect censorship and signs of
  surveillance
 
-[] OONI software tests allow users to circumvent censorship
+[ ] OONI software tests allow users to circumvent censorship
 
-[] OONI software tests allow users to circumvent censorship and surveillance
-
-## Which of the following statements is true?
-
-[] ooniprobe is run over Tor, thus protecting my privacy
-
-[] ooniprobe is not run over Tor, but is still designed to protect my privacy
-
-[] ooniprobe is not run over Tor, but measurements can get sent over Tor
-
-[] ooniprobe measurements are sent over Tor, thus protecting my privacy
+[ ] OONI software tests allow users to circumvent censorship and surveillance
 
 ## Which of the following statements is true?
 
-[] Anyone monitoring my internet activity (e.g. ISP, government or employer)
+[ ] ooniprobe is run over Tor, thus protecting my privacy
+
+[ ] ooniprobe is not run over Tor, but is still designed to protect my privacy
+
+[ ] ooniprobe is not run over Tor, but measurements can get sent over Tor
+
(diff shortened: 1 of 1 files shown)


informed-consent/quiz.md
```diff
@@ -0,0 +1,29 @@
...
+## is true?
+
+[ ] Anyone monitoring my internet activity (e.g. ISP, government or employer)
+ might be able to see that I am running ooniprobe, even though OONI takes
+ precautions to make this hard
+
+[ ] Anyone monitoring my internet activity (e.g. ISP, government or employer)
+ will *not* be able to see that I am running ooniprobe
+
+[] ooniprobe is designed to protect my privacy and therefore my use of ooniprobe
```
bassosimone: nit here: [] => [ ]

informed-consent/quiz.md
```diff
@@ -0,0 +1,29 @@
...
+ cannot be detected
+
+
+## In regards to the publication of measurements, which of the following
+## statements is true?
+
+[ ] My measurements will *not* by default get published on OONI Explorer
+
+[ ] My measurements will by default get published on OONI  Explorer and might
```
bassosimone: nit: two spaces after OONI and before Explorer

bassosimone:
1. Would it also make sense to have a question related to the risk of running specific tests? For example, to see whether the user understand that http-invalid-request-line entails more risks than running the web-connectivity test?

2. If I'm understanding correctly the scenario, would it make sense to check whether the user has clear that, in addition to measurements explicitly requested, lepidopter (and also ooni-probe when there is orchestration) will run measurements in the background?

3. I feel like there should be also a question on the (soft or hard) legality of running ooni-probe

In general, I think the objective of the quizzes would be to make sure users have clear at least the points spelled out in the inform-user-short document, isn't it?

hellais:
re: nits, the markdown in here is actually not going to be parsed and used for generating the HTML in the web UI. The actual stuff that will end up in the GUI can be found here: 
https://github.com/TheTorProject/ooni-wui/blob/master/src/app/components/setup/step1.html
https://github.com/TheTorProject/ooni-wui/blob/master/src/app/components/setup/step2.html
https://github.com/TheTorProject/ooni-wui/blob/master/src/app/components/setup/step3.html

hellais:
@agrabeli could you perhaps push these changes directly to the ooni-wui repository or open tickets with the proposed changes in there so that it's easier for me to integrate the feedback there?

The Quiz is not being generated from the markdown in this pull requests so it's going to be hard to sync it up if commits are pushed on top of this branch.

I would suggest this PR Is closed and we discuss changes as either tickets in ooni-wui or as pull requests there.

hellais:
Closing as this is now superseeded by the copy tracked on transifex and in the apps.

PR #76, opened 2016-10-12
Add one-call API for submitting one-entry reports

bassosimone:
Discussing with partners interested to use our collector, I noticed that, when there is need to submit a single-entry report, it would be a better selling point telling them "look, use this simple API and you're set". For this reason, I decided to specify this simple API and I am willing to implement it into ooni-backend if it is approved.


oonib.md
```diff
@@ -256,6 +263,60 @@ running the test and the report can be considered done:
 
 `POST /report/$report_id/close`
 
+### 2.3.1.4 Submitting simple reports using a single API call
```
hellais: An alternative could be to "un-deprecate" the "content" key inside of the `/report` API endpoint. I would then put all the required keys inside of the top level JSON structure and the "content" key would become "test_keys"

bassosimone: True. I spent some time thinking at both alternatives. I believe it's more simple to communicate to people and more straightforward design to have a separate API serve a separate purpose. My take is that it's easier to imagine routing being done at the level of the URL rather than by adding or not adding a key into the data. Just think at how simpler it is to describe (and search for) this functionality in the current proposal (where you can say that upfront) compared to having to document in the initial paragraph that by changing a bit of data into a sub part of the specification (2.3.1.1 if I understand correctly) you can get different behavior.

What do you think? Does this make sense?

hellais:
I have actually implemented something close to this inside of the next generation collector: https://github.com/ooni/collector/blob/master/collector/api/v1/api.go#L48.

I am going to update this PR with that proposed API and suggest we merge this.

oonib.md
```diff
@@ -1,7 +1,7 @@
 # oonib specification
 
-* version: 1.3.0
-* date: 2016-02-05
+* version: 1.4.0
+* date: 2016-10-12
```
bassosimone: 2018-10-12?

oonib.md
```diff
@@ -256,6 +263,60 @@ running the test and the report can be considered done:
...
+
+When the request is successful the backend shall return:
+
+`Status code: 200 (OK)`
+
+Message:
+
+```JSON
+{
+"report_id": "XXXX"
```
bassosimone: Comma missing here

hellais:
I am going to merge this.

PR #78, opened 2016-11-25
Remove informed-consent and test-descriptions from the ooni-spec repo

hellais:
Their new canonical home is the ooni-web repository inside of the
following directories:

    * ooni-web/content/nettest/
    * ooni-web/content/about/


anadahz:
Looks good merging.

PR #81, opened 2017-01-09
Feature/opos

hellais:
This branch include work in progress on the probe orchestration system


opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+* The start time of the job in ISO 8601 (YYYY-MM-DDThh:mm:ssTZ, where TZ is the timezone of `Z` for UTC).
+  An empty start time means start immediately.
+
+* The run interval, defined following the ["Duration" component of the ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) standard.
+
+The Management Interface will assign to a given job a unique id (hereafter
+referred to as `job_id`) that can be used to unschedule jobs.
+
+It should be possible for somebody interacting with the Management Interface to
```
bassosimone: Here I'd say "an authorised operator"

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+schedule the following types of actions.
+
+### 4.1.1 Run test
+
+This action is used to instruct a probe to run a certain OONI test once. For
+sake of clarity we omit the common elements from the above format.
```
bassosimone: Suggestion: can we design a format by which we *can* omit common fields unless necessary to override some defaults, such that the JSON that will travel on the wire is actually the one in here?

bassosimone: Unrelated: It would help to clarify in some way from which two entities will this message travel.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+![OPOS Job Lifecycle](http://g.gravizo.com/g?@startuml;
+    [*] --> READY;
+    READY --> ASSIGNED;
+    ASSIGNED --> STARTING;
+    ASSIGNED --> REJECTED;
+    STARTING --> RUNNING;
+    RUNNING --> COMPLETED;
+    COMPLETED --> [*];
+    @enduml
+)
```
bassosimone: Is this a picture?

bassosimone: Ah, never mind: it is a picture (I just checked by viewing the document)

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+* **RUNNING**: when the job has started it enters this state. The OPOS Client
+  is responsible for reporting back to the **OPOS Scheduler** updates on the
+  progress of any given task.
+
+* **COMPLETED**: when a job has finished executing it enters the COMPLETED state.
+
+
+# Links
+
+These
```
bassosimone: I think we can omit this "these"

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+**OPOS Client**: This is generally an instance of ooniprobe (or
+ooniprobe-mobile) that is subscribing to the OPOC event feed.
+
+**OPOS Scheduler**: This is the component that knows the status of the jobs
+scheduled via the OPOS Management Interface, which OPOS Clients have taken them
+on and their progress.
+
+**OPOS Event feed**: This is the backend component responsible for emitting
+events to clients and informing them of what they should be doing.
```
hellais: As a matter of fact the "scheduler" was added after the "event feed". Originally I had excluded the "scheduler" from the architecture as I was interested in documenting only what is exposed and the "scheduler" in a way is an architectural construct that is in a way transparent to the user.
However in specifying this I felt the need to have a this architectural construct and so I added it.

To exemplify this in a more concrete way, a possible implementation of this would be:

* OPOS Event feed: a websockets interface that clients subscribe to receive notices of what they should do
* OPOS Scheduler: the engine responsible for knowing which clients should be notified of events

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+* It MUST be possible to signal the **OPOS Client** when a new event that
+  interests them is scheduled.
+
+* It SHOULD be possible to place the **OPOS Event feed** behind a domain front
+  or leverage some other collateral freedom like principle (for example hosting
+  it on github.com) to avoid it being blocked hence being resilient to censorship.
+
+* It SHOULD be possible to know how many *active* **OPOS Clients** there are at
+  a given time.
```
hellais: That is why I mentioned this as a SHOULD and not as a MUST. It's something I would like to have the power to know, because it's very useful, but it's not a strict requirement.

I agree that depending on the specific implementation there will be different ways to do this that can be more or less accurate. Worst case scenario we can do something similar to how Tor counts users. We estimate the average number of requests each client will make per day and count the number of requests and divide by the number of average requests a client is meant to make.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+                "or": [
+                    {"in": ["IT", "NZ"]},
+                    {"limit": 10, "in": ["US"]}
+                ]
+            },
+            "probe_asn": "",
+            "probe_id": "",
+            "probe_family": "",
+
+            "platform": "macos", # Can be one of lepidopter, macos, linux, mobile, android, ios
```
hellais: I guess I should have made it more clear, but actually any key can be a filter definition so you can say:

```
"platform": "macos"
```

OR 
```
"platform": ["macos", "ios"]
```

OR

```
"platform": {
  "or": [
    {"in": ["macos", "ios"]},
    {"limit": 10, "in": ["android"]}
  ]
}
```

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+            "network_type": "wifi" # can be one of wifi, mobile, 3g, 4g, edge
```
hellais: Same as above. I will add more details about the filter system we plan to support.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+    }
+}
+```
+
+The where clause in the filter definition is an implementation of the [loopback
+style filters](https://github.com/strongloop/loopback-filters), but with "in"
+in place of "inq".
+
+run now contains a boolean value that if true means this client should execute
+the action, false if it should not.
```
hellais: I think this is some leftover text from a previous iteration of the document. Will remove.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+`delay`: is an upper bound on the number of seconds to delay the action by. For
+example by specifying `"delay": 60` the action will be triggerred after a
+number of seconds going from `0-60`.
```
hellais: Yeah sorry, I did not update the above reference to `random_delay`. I suggest we call it always `delay` even if it's actually random. The reason for doing so is that if you want to delay a task by some precise value you should just schedule it at the desired time and not use delay.

I have found this to be consistent with the naming scheme of other scheduling systems (ex. see advanced section of https://technet.microsoft.com/en-us/library/cc748841(v=ws.11).aspx)

opos/OONI-Probe-Orchestration-System-Design.md
hellais: Yeah it's a UML chart.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,286 @@
...
+
+**OPOS Event log**: Through this it is possible to inspect the history of
+instrumented events.
+
+A **OPOS Client** will subscribe to the **OPOS Event feed** to receive
+information on what it should do.
+
+When somebody with proper access permissions adds an event to via the **OPOS
+Management Interface** all the subscribed probes that are affected by that
+event will be notified and will begin performing the specified action.
```
hellais: I added a note about this below.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+time a new instruction is inserted into the OPOS to the moment that the
+affected probes receive it 1 minute should pass.
+
+## 1.3 Granular
+
+It should be possible to give instructions based on:
+
+* The country of a probe
+
+* The Network of a probe
```
bassosimone: bikeshedding: I don't think we should capitalize network here

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+instrumented events.
+
+A **OPOS Client** will subscribe to the **OPOS Event feed** to receive
+information on what it should do.
+
+When somebody with proper access permissions adds an event to via the **OPOS
+Management Interface** all the subscribed probes that are affected by that
+event will be notified and will begin performing the specified action.
+
+The **OPOS Scheudler** is the scheduling engine responsible for figuring out
```
bassosimone: Typo here

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+which clients to notify of jobs and how.
+
+A user interested in seeing what has been scheduled in the past can visit the
+**OPOS Event log** and see a history of what has happenned.
+
+# 3.0 Transport protocol requirements
+
+We use the term "Transport" loosely here to mean the protocol that is being
```
bassosimone: I don't think Transport should be capitalised here (and let's paint the bike shed ooni-blue)

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+affected resources, to this end it's also possible to delay the action by a random delay
+by specifying the optional `delay` key.
+
+We use JSON serialization for representing the message structures inside of
+this document, however if due to other contraints JSON is not an adequate
+serialization protocol we are not necessarily tied to it.
+
+The base data format is the following:
+
+```json
```
bassosimone: bikeshedding: if we declare this as json, the comment (which is not json) will result badly coloured. I think it's better either to have a valid json or to do not declare this as json.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+        "where": {
+            "probe_cc": {
+                "or": [
+                    {"in": ["IT", "NZ"]},
+                    {"limit": 10, "in": ["US"]}
+                ]
+            },
+            "probe_asn": "",
+            "probe_id": "",
+            "probe_family": "",
```
bassosimone: What is this family?

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+            "url": "https://torproject.org/"
+        }
+    }
+}
+```
+
+`filter`:
+The `where` clause in the filter definition is an implementation of the [loopback
+style filters](https://github.com/strongloop/loopback-filters), but with "in"
+in place of "inq" for clarity.
```
bassosimone: I know I originally advocated for `in` rather than `inq`. Now, I'm wondering whether keeping `inq` will be better because we adhere to the standard and we have one less diff-y thing to remember.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+  scheduler of such and the task state becomes **REJECTED**.
+
+* **RUNNING**: when the job has started it enters this state. The OPOS Client
+  is responsible for reporting back to the **OPOS Scheduler** updates on the
+  progress of any given task.
+
+* **COMPLETED**: when a job has finished executing it enters the COMPLETED state.
+
+
+# Links
```
bassosimone: I'd perhaps link inline rather than at the bottom, so it's more clear what refers to what

opos/Measurements-and-url-policy.md
```diff
@@ -0,0 +1,43 @@
+# Measurements and URL policy
+
+This document explains the policy we follow when considering what URLs and
+Measurements should be scheduled via OPOS and/or any other system run by OONI
+to gather network measurement data.
+
+1. To the extent that it is possible we will always do what is best for users of OONI.
```
anadahz: I guess it should be mentioned to the policy that users can be opt-out by OPOS and will not be enabled by default.

opos/Measurements-and-url-policy.md
```diff
@@ -0,0 +1,43 @@
+# Measurements and URL policy
+
+This document explains the policy we follow when considering what URLs and
+Measurements should be scheduled via OPOS and/or any other system run by OONI
```
anadahz: The OPOS initials could be expanded here.

opos/Measurements-and-url-policy.md
```diff
@@ -0,0 +1,43 @@
...
+
+5. A URL must first have been included as part of the
+   [citizen-lab testing list](https://github.com/citizenlab/test-lists) repository to be
+   considered for testing.
+
+6. Only people in the governance roster can trigger measurements and a
+   measurement must be signed off by at least one other person.
+
+7. Anybody in the governance roster can nominate somebody else to join the
+   roster. Their inclusion shall be discussed and once there is consensus they
```
anadahz: I guess the inclusion of people or entities *must* (and not shall) be discussed beforehand.

opos/Measurements-and-url-policy.md
```diff
@@ -0,0 +1,43 @@
...
+   are added. To be included in the roster you need to:
+
+   a. Have a legittimate reason to do so (ex. you do research on censorship)
+
+   b. Be a respected member of the OONI community
```
anadahz: We should be very careful with people using the OPOS not only for (obvious) reasons of malice but also for reason of experimentation failure: eg: researching on a new censorship analysis that could de-anonymize users.

I assume that the bigger this governance roster became the harder will be to control OPOS usage.

opos/Measurements-and-url-policy.md
hellais: I think that this is a bit out of scope for the policy document as this defined the governance and rules for when OPOS is in place. Whether it is in place for all users, opt-in, disableable or whatnot is an implementation detail.

BTW I think we should have it enabled by default for all **new** users and prompt old users to enable it.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+
+Yet, it is still important that a probe can properly function even when the
+OPOS is unreachable.
+
+## 1.2 Latency
+
+The target latency for the OPOS is of around 1 minute. This means that from the
+time a new instruction is inserted into the OPOS to the moment that the
+affected probes receive it 1 minute should pass.
+
```
anadahz: I found the latency window quite short, is there any reasoning for this latency?

opos/Measurements-and-url-policy.md
hellais: I think shall and must are synonyms in this context, though shall sounds to me as a bit softer.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: Yeah that was also my reasoning for using `inq` in the beginning. I don't have a strong opinion for either (maybe we can say that they are both supported?)

opos/OONI-Probe-Orchestration-System-Design.md
hellais: I guess it's a bit confusing to include something that is not yet implemented in ooniprobe here, but my reasoning behind it was that at some point we may want to group ooniprobes together by family, so that you can say "the ooniprobe run by ORGX" and they are 10 each with their own ID.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: Yes you are right, I guess I should declare it as `hjson` (https://hjson.org/)

opos/OONI-Probe-Orchestration-System-Design.md
hellais: By short you mean it's too little? 1 minute of latency is actually fairly high, even with long polling you are going to achieve latencies that are less than 1 minute.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+## 1.3 Granular
+
+It should be possible to give instructions based on:
+
+* The country of a probe
+
+* The Network of a probe
+
```
anadahz: Some useful granular items that could be added:
* Installed tests
List of installed tests per probe.
* Allowed tests
List of permitted tests to be run per probe.
* Type of collector (onion, https, cloudfront) used to submit reports
* Privacy options enabled per probe (includeip, includeasn, includecountry)
* Size of measurement quota (limit, free %)

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+When somebody with proper access permissions adds an event to via the **OPOS
+Management Interface** all the subscribed probes that are affected by that
+event will be notified and will begin performing the specified action.
+
+The **OPOS Scheudler** is the scheduling engine responsible for figuring out
+which clients to notify of jobs and how.
+
+A user interested in seeing what has been scheduled in the past can visit the
+**OPOS Event log** and see a history of what has happenned.
+
```
anadahz: A note should be added with the default log retention date.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+style filters](https://github.com/strongloop/loopback-filters), but with "in"
+in place of "inq" for clarity.
+Any of the sub-keys inside of the "where" keys support the filters. This means
+you can say: `"platform": "macos"` or `"platform": ["macos", "android"]`.
+
+`delay`: is an upper bound on the number of seconds to delay the action by. For
+example by specifying `"delay": 60` the action will be triggerred after a
+number of seconds going from `0-60`.
+
+`schedule`: The scheduling for the job, in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format. Consists of 3 parts separated by /:
```
anadahz: extra `/`

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+{
+    "schedule": "R//P1D",
+    "task": {
+        "name": "Check for accessibility of https://torproject.org/",
+        "ooni": {
+            "test_name": "web_connectivity",
+            "url": "https://torproject.org/"
+        }
+    }
+}
```
anadahz: Shouldn't ' a `job_id` be set to schedule a test?

opos/Measurements-and-url-policy.md
bassosimone: > BTW I think we should have it enabled by default for all new users and prompt old users to enable it.

Yes, I agree with this

opos/OONI-Probe-Orchestration-System-Design.md
hellais: I don't think these should go inside of the high-level design goals. The purpose of this section is to set the expectations as to what broad type of problem this system is trying to solve (schedule measurements based on network and location of a probe), the specific requirements (that are due to implementation details) that are needed in order to satisfy the job is too low level for at least this section at least.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: There is no log retention date. We should store all logs forever.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: No, as is mentioned below:

>The Management Interface will assign to a given job a unique id (hereafter referred to as `job_id`) that can be used to unschedule jobs.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: The `/` character is the separator for the 3 parts. I should probably wrap it in ``` though

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+    RUNNING --> COMPLETED;
+    COMPLETED --> [*];
+    @enduml
+)
+
+
+The OPOS Scheduler keeps track of the lifecycle of a job. With respect to an
+**OPOS Client** a job can be in one of the following states:
+
+* **READY**: this is the state in which a job is when it is first added via the OPOS Management Interface.
```
anadahz: For consistency this can be: `this is the state a job is in when[...]`

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+* **REJECTED**: if an OPOS Client cannot run a certain job (either because of
+  some the constraints not being satisfied or something else) it informs the
+  scheduler of such and the task state becomes **REJECTED**.
+
+* **RUNNING**: when the job has started it enters this state. The OPOS Client
+  is responsible for reporting back to the **OPOS Scheduler** updates on the
+  progress of any given task.
+
+* **COMPLETED**: when a job has finished executing it enters the COMPLETED state.
+
```
anadahz: For consistency the REJECTED RUNNING COMPETED states can start with: `this is the state a job is in when[...]`

opos/Measurements-and-url-policy.md
```diff
@@ -0,0 +1,43 @@
...
+   considered for testing.
+
+6. Only people in the governance roster can trigger measurements and a
+   measurement must be signed off by at least one other person.
+
+7. Anybody in the governance roster can nominate somebody else to join the
+   roster. Their inclusion shall be discussed and once there is consensus they
+   are added. To be included in the roster you need to:
+
+   a. Have a legittimate reason to do so (ex. you do research on censorship)
```
anadahz: Typo: s/legittimate/legitimate

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+When somebody with proper access permissions adds an event to via the **OPOS
+Management Interface** all the subscribed probes that are affected by that
+event will be notified and will begin performing the specified action.
+
+The **OPOS Scheudler** is the scheduling engine responsible for figuring out
+which clients to notify of jobs and how.
+
+A user interested in seeing what has been scheduled in the past can visit the
+**OPOS Event log** and see a history of what has happenned.
```
anadahz: Typo: s/happenned/happened

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+When no target is specified the action will affect all clients that support the
+specified action.
+
+It is sometimes desirable to avoid having all probes run the specific action at
+the same time, for example because it could lead to a denial of service of the
+affected resources, to this end it's also possible to delay the action by a random delay
+by specifying the optional `delay` key.
+
+We use JSON serialization for representing the message structures inside of
+this document, however if due to other contraints JSON is not an adequate
```
anadahz: Typo: s/contraints/constraints

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+
+`filter`:
+The `where` clause in the filter definition is an implementation of the [loopback
+style filters](https://github.com/strongloop/loopback-filters), but with "in"
+in place of "inq" for clarity.
+Any of the sub-keys inside of the "where" keys support the filters. This means
+you can say: `"platform": "macos"` or `"platform": ["macos", "android"]`.
+
+`delay`: is an upper bound on the number of seconds to delay the action by. For
+example by specifying `"delay": 60` the action will be triggerred after a
```
anadahz: Typo: s/triggerred/triggered

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+* The number of times to repeat the job: Rn to repeat n times, or R to repeat forever
+
+* The start time of the job in ISO 8601 (YYYY-MM-DDThh:mm:ssTZ, where TZ is the timezone of `Z` for UTC).
+  An empty start time means start immediately taking into account the eventual
+  random delay when the `delay` key is specified.
+
+* The run interval, defined following the ["Duration" component of the ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) standard.
+  Durations are indicated as a string in the format
+  `P[n]Y[n]M[n]DT[n]H[n]M[n]S`, where `[n]` is a number representing a value
+  for the follwoing date or time element.
```
anadahz: Typo: s/follwoing/following

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: By short I mean too little, given the fact that there are a number of probes on very slow links getting a new instruction per minute can be too much.

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: These are some important properties that need to be enumerated in advance otherwise we may "violate" some user's options. 

Some possible scenarios:
- Schedule tests that are not installed by probes.
- Schedule tests that are not "allowed" to run ie: the http header field manipulation test.
- Schedule a test with a long list of experiments that require a decent amount of disk space to probes reaching their disk quota capacity.

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: There are a number of probes that cannot handle storing all logs forever.
Perhaps it make sense to add a default log retention plan rather than end up having probes running out of disk space.

hellais: > There are a number of probes that cannot handle storing all logs forever.

The logs are not stored by the probes, but by the probe orchestration system.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: Even if the link is slow (in terms of throughput), they will most likely have a latency < 60 seconds.
If they do have a latency > 60 seconds TCP sockets will timeout before they are established and bumping the time up will not help much there.
In which circumstances do you imagine there to be a > 60 seconds latency with OPOS?

opos/OONI-Probe-Orchestration-System-Design.md
hellais: > Some possible scenarios:
> Schedule tests that are not installed by probes.
> Schedule tests that are not "allowed" to run ie: the http header field manipulation test.
> Schedule a test with a long list of experiments that require a decent amount of disk space to probes reaching their disk quota capacity.

I think these are implementation details that don't necessarily have to live in this document, or at least not in this section. As an operator I don't care to specify these options and the orchestrator should be smart enough to figure out which of the available probes is capable of running my measurements.

I can add a note about the fact that certain attributes of the probe need to be sent to the OPOS in order for it to be able to resolve these constraints, in `4.0 Jobs`.

opos/OONI-Probe-Orchestration-System-Design.md
```diff
@@ -0,0 +1,302 @@
...
+* Run this instruction on at most `N` probes inside of network `K` of country
+  `C`.
+
+## 1.4 Transparent
+
+There should be full transparency over what is being scheduled to be run on
+orchestrated probes. That is to say that all measurements scheduled via OPOS
+must be recorded in an append only log so that it's possible, upon request,
+for a user to learn what jobs have affected their probe.
+
```
anadahz: How and from which entity a probe-user can request backlog of the past scheduled tasks?

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: For privacy reasons it would be better if these logs are stored only on the probe(s) that performed the scheduled tests and are not stored in a service online.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: > **OPOS Event log**: Through this it is possible to inspect the history of
instrumented events.

opos/OONI-Probe-Orchestration-System-Design.md
hellais: > For privacy reasons it would be better if these logs are stored only on the probe(s) that performed the scheduled tests and are not stored in a service online.

but as you pointed out before probes can have limited disk space available and they can't store them all.

Moreover there is nothing stopping somebody from subscribing to the event feed as all possible probe types and logging all the messages.
I think the assumption is that it's safe to log all the scheduled events as they are being broadcasted publicly.

What exactly is the privacy implication of making the scheduled events be known public?

opos/OONI-Probe-Orchestration-System-Design.md
hellais: I am just going to drop the links that are not mapped to anything in particular. The reason why I put chronos there originally was for part of the inspiration of the scheduler format.

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: I imagine that in most cases the probes will not have completed the scheduled job in less than 60 seconds. I can't think of many good reasons of having such a short polling.

hellais: The 1 minute latency is from when they **receive** it, that doesn't mean it will take them less than 1 minutes to **complete** it.

hellais: Anyways I don't want to fixate on this as in the end this is going to be an implementation detail. The purpose of specifying this is that the order of magnitude we are targeting is not seconds, but rather minutes and not hours.

Would it make you more comfortable if I were to say "minutes" instead of "1 minute"?

bassosimone: Yeah, minutes rather than one minute is reasonable as well. In any case, the probe ability to refuse running a job is IMO our best line of defense against over scheduling.

hellais:
I think all feedback should have been addressed. Can I proceed with merging this?

bassosimone:
Hit it!

opos/OONI-Probe-Orchestration-System-Design.md
anadahz: Sure that makes and it's definetely not a blocker for the OPOS spec.
Thanks for working on this @hellais .

PR #87, opened 2017-06-15
Update captive-portal spec

joelanders:
I looked around for what Microsoft, Google, and Apple do for captive portal detection these days and wrote a test for it in measurement-kit here:
https://github.com/measurement-kit/measurement-kit/pull/1238

I'm proposing a different report format here.

### DIFF
--- a/test-specs/ts-010-captive-portal.md
+++ b/test-specs/ts-010-captive-portal.md
@@ -25,7 +25,7 @@ The test is split in four smaller tests:
     3. DNS test that compares an A request to a static domain against its IP (part of the vendor_dns_tests).
     4. DNS test that check for 0x20 encoding (check0x20 tests).
 
-## Version 0.4.0 (currently only in measurement-kit):
+## Version 0.4.0 (implemented in MK, but not deployed):
 
 Most of the vendor tests (from Microsoft, Apple, and Google) do an HTTP request and verify the status code or body content matches some expected value. One of the tests (from Microsoft) only checks that a DNS lookup returns the correct IP address.
 
--- a/test-specs/ts-010-captive-portal.md
+++ b/test-specs/ts-010-captive-portal.md
@@ -29,6 +29,31 @@ The test is split in four smaller tests:
 
 Most of the vendor tests (from Microsoft, Apple, and Google) do an HTTP request and verify the status code or body content matches some expected value. One of the tests (from Microsoft) only checks that a DNS lookup returns the correct IP address.
 
+### References
+#### MS tests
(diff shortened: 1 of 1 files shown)


bassosimone:
I am mostly okay with these changes. What I am wondering is whether it would perhaps be possible to keep similar keys to the previous version, so to have a common parser that is able to parse either version 0.3 or version 0.4 (equal to 0.3 but with more keys).

Another aspect that I think should be emphasized is that you say you have checked what is currently being done *nowadays* by the vendors. Since this seems to be different from what we were doing previously, I was thinking that perhaps you should document the sources from which you did ready what they do, so they are here for the records.

I'd also like to see whether @hellais, also pinged [here](https://github.com/measurement-kit/measurement-kit/pull/1272#issuecomment-313447783) has further comments.

hellais:
Why was this PR closed?

hellais:
I think the main point of friction is whether or not we can preserve some minimal amount of backward compatibility for the test or if it's necessary to just scratch everything and re-do it from scratch.

joelanders:
Ah, I was going to re-open another PR here that just added citations for the methodology. I think we can keep the report format (working on that now).

joelanders:
Ok, never mind, some key names have to be added, so let's keep this PR open and I'll modify it.

willscott:
@bassosimone is this reflecting the current state of MK (the v0.4 of the spec in this PR claims to explain MK behavior)? if so, we should merge this

test-specs/ts-010-captive-portal.md
```diff
@@ -16,13 +16,44 @@ The ability to detect a captive portal.
...
+## Version 0.3.0 and earlier:
+
 The test is split in four smaller tests:
 
     1. Http test that compares a requested document against the control one (vendor_tests).
     2. DNS test that checks a NXDOMAIN response to random hostnames requests (part of the vendor_dns_tests).
     3. DNS test that compares an A request to a static domain against its IP (part of the vendor_dns_tests).
     4. DNS test that check for 0x20 encoding (check0x20 tests).
 
+## Version 0.4.0 (currently only in measurement-kit):
```
hellais: I believe this is currently not implemented. Probably we should change this bit of text and say that this is work in progress.

hellais: Actually this is not true. However we are not running it from the mobile app and it's still missing some features imho. I documented those here: https://github.com/measurement-kit/measurement-kit/issues/1577

hellais:
I updated the section header and am merging this.

PR #88, opened 2017-06-16
WIP: Schema validator

endorama:
Following a hacknight with @bassosimone in Turin, I started working on the JSON schema validator ( https://github.com/TheTorProject/ooni-probe/issues/484 ).

This is first take at it, please consider as a WIP still to be refined.

The idea is to use [JSON schema](http://json-schema.org/) to write down a machine-usable version of the test-specs in this repository, in order to automatically perform validation.

Currently only a validator for report header is present ( and properly working ).

Tested using python 3.5.2 ( mainly due to better unicode support and nicer output ). Test are run using [jsonschema](https://pypi.python.org/pypi/jsonschema) python package from command line, but can be completely implemented in python.

You can test each file inside `schema-validator/test-specs` using:

```
$ cd schema-validator
$ python validators/header.py -s schemas/header.json test-specs/ts-002-http-host.json
```

I extrated all JSON examples from test-specs, and put in `schema-validator/test-specs`. In the future, when the JSON schema is complete for all reorts, test-specs could be generated directly from it.

Further steps discussed with @bassosimone :
- complete report header JSON schema
- add specific report test JSON schemas
- convert schema file from JSON to YAML for readability ( the validator script will convert them back to JSON )

Some examples in test-specs are still in YAML format. I ignored those files because it appears that the specification is old, but please correct me if I'm wrong :smiley: 

Opinions/suggestions more than welcome.


endorama:
@bassosimone I'm updating the PR with:

- some documentation! See README.md
- YAML instead of JSON for writing JSON schema
- validator script is a bit more robuts ( and should be self explanatory now )

Feedbacks welcome!

bassosimone:
@endorama Thanks! Sadly, I am a bit swamped until Monday, so I cannot dedicate enough time to do a proper review. Will do that next week. For now, cc-ing @hellais and @joelanders with whom I had discussed this feature previously and that most certainly are interested to see the progress.

schema-validator/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Folder structure
+
+```
+├── requirements.txt 	required python libraries
+├── schemas 			JSON Schemas
+├── test-specs 			JSON representation of markdown test-specs
+├── validate-all.sh 	utility script for validating all test-specs against
+│ 							header schema and specific test schema
+└── validate.py 		validator script ( run with -h for help )
```
bassosimone: Bike shedding: it seems the lines are not aligned (maybe a space vs. tab issue?)

schema-validator/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## JSON Schema
```
bassosimone: I think here you want to close the three backticks, otherwise the whole readme has the verbatim style.

schema-validator/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+> JSON Schema is a vocabulary that allows you to annotate and validate JSON documents.
+
+Quoted from [json-schema.org][json-schema-website]
```
bassosimone: Shouldn't this be `[text](URL)`? [Edit: ah, no, stupid me, this is an indirect link]

schema-validator/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+An useful tool for generating/editing JSON schema via a graphical editor could
+be found [here][jsonschema.net].
+
+## How to validate
+
+Validating JSON Report file can be done using `validate.py`: pass to the python
+script the schema to be used for validation and the JSON file to validate.
+
+For proper CLI instruction run `validate.py -h`.
```
bassosimone: Arrived here, I stopped and started thinking that probably what would be cool is to have this package in `pip` along with the schema files, somehow, such that one, as part of running regress tests could `pip install $this_package_name` and then invoke the validator on the produced reports. @hellais, what do you think?

schema-validator/README.md
```diff
@@ -0,0 +1,49 @@
...
+## TODO
+
+- Add specific test schemas: currently the only schema available is the Header
+  schema. Specific schema tests are missing.
+- Gather the proper format for YAML test-specs: some specifications are still
+  in YAML format. They need to be ported to JSON.
+- Produce example JSON from JSON schemas: each file in the test-specs folder
+  contains an "example" paragraph with a JSON example of the Report payload.
+  This payload could be generated directly from JSON schemas. A nice tool for
+  the work could be [json-schema-faker][json-schema-faker]
```
bassosimone: Oh, this is epic and super interesting! (Side note: perhaps it would even be fine, in the interest of removing duplication, to tell people that one can get an example of them schema of test `x` by running `$tool_name --example x`. Again, here, @hellais, what do you think?)

schema-validator/schemas/header.yml
```diff
@@ -0,0 +1,90 @@
...
+definitions:
+  bucket_date:
+    type: string
+    pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
+  semantic_version:
+    type: string
+    pattern: "^[0-9]+.[0-9]+.[0-9]+$"
+  uuid-v4:
+    type: string
+    pattern: "[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\\Z"
```
bassosimone: (Maybe this question is caused by my ignorance, anyway) what is the reason to quote `Z` here?

schema-validator/schemas/header.yml
```diff
@@ -0,0 +1,90 @@
...
+    anyOf:
+      - type: number
+        minimum: 0
+      - type: "null"
+  test_start_time:
+    $ref: "#/definitions/date-time"
+  test_version:
+    $ref: "#/definitions/test-version"
+required:
+  - bucket_date
```
bassosimone: I remember this being part of the specification, but I'm 99% sure this is something added by the backend pipeline, and, as such, it should not be mandatory (measurement-kit and ooni-probe does not generate it).

schema-validator/schemas/header.yml
```diff
@@ -0,0 +1,90 @@
+---
```
bassosimone: @endorama so, this is basically the specification for the header rewritten in YAML, correct? I think this is much more readable than when I saw it in JSON, great!

schema-validator/validate.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
bassosimone: @endorama can you run `autopep8` on this file?

schema-validator/validate.py
```diff
@@ -0,0 +1,62 @@
...
+			print("{} is not a valid target file. Only JSON files allowed".format(args.file))
+			sys.exit(1)
+
+	if not os.path.isfile(args.schema):
+		print("Schema file {} does not exists".format(args.schema))
+		sys.exit(1)
+	else:
+		if not args.schema.endswith('yml'):
+			print("{} is not a valid schema file. Only YAML files allowed".format(args.schema))
+			sys.exit(1)
```
bassosimone: Perhaps it's a bit redundant to have `else` after `sys.exit` (but it's probably a matter of taste and I recognize that my bikeshedding index is reaching 2,048 after I wrote this sentence :-P)

bassosimone:
So, it seems I had some time to do some bikeshedding (actually, I was curious and couldn't resist peeking a bit into this). I guess the biggest question is whether we want to make this something that is on the package index and can be downloaded with `pip` so we can validate reports as part of regress tests.

schema-validator/schemas/header.yml
endorama: I remember I was having problems without it, and a post somewhere on the internet suggested to do so. But I can't find it anymore, so I'm removing it.

schema-validator/schemas/header.yml
endorama: Yes it is. Is a JSON-Schema file in YAML instead of JSON. I also prefer it this way.

schema-validator/validate.py
endorama: Done!

endorama:
I'm sorry for the long delay, following @bassosimone comments I updated the PR.

hellais:
I am also very sorry for taking ages to take a look at this. I started working on some things that are somewhat related to this, so I am going to take on merging the work that was done as part of this into another branch and eventually integrate it into master.

Overall this looks very good to me. @endorama thank you so much for working on it!

hellais:
I will add as notes to this PR the things that I encounter as I review it.

schema-validator/requirements.txt
```diff
@@ -0,0 +1 @@
+jsonschema
```
hellais: `pyyaml` is also a dependency.

schema-validator/schemas/header.yml
```diff
@@ -0,0 +1,89 @@
...
+    pattern: "[0-9]{4}-[0-9]{1,2}-[0-9]{1,2} [0-9]{2}:[0-9]{2}:[0-9]{2}"
+properties:
+  bucket_date:
+    $ref: "#/definitions/bucket_date"
+  data_format_version:
+    $ref: "#/definitions/semantic_version"
+  id:
+    $ref: "#/definitions/uuid-v4"
+  input:
+    $ref: "#/definitions/ip-address"
```
hellais: `input` seems to be defined as a `ipv4` address, which is actually not the case, however the validation says it succeeds to validate `test-specs/ts-017-web-connectivity.json`, which has as `input`: `http://torproject.org/`, which is clearly not a ipv4 address.

So either there is a problem in the validation logic or in the definition of what is an ipv4 address.

hellais:
I pushed my changes to this branch: https://github.com/ooni/spec/tree/schema-validator

endorama:
Hello @hellais, thanks for stepping in. 

Please let me know if I can help in any way, I would gladly contribute again to the project.

hellais:
I picked this task up again yet took the direction of porting it to an avro spec.

The benefit of defining an avro spec is that it's the first step necessary to offer an alternative way to download measurements in parquet or avro (which are very commonly used in big data tooling).

I pushed all the code to the above branch and the next steps I see are:

* [ ] Fixup the schemas for the legacy tests
* [ ] Sample more real measurement and check that the schemas will pass validation (I only sampled a single test from every test class)
* [ ] Run this against a larger selection of the OONI dataset and check that it still works as expected
* [ ] (maybe) Integrate part of this into the pipeline

bassosimone:
We have struggled to complete this PR for a long time now. I believe this is an indication that we were probably attacking a too complex problem. I have seen code in the OONI pipeline written by @FedericoCeratto and code in github.com/ooni/jafar that give us enough confidence that the data is well formed, that I guess we can close this PR at this point.

PR #95, opened 2017-10-24
Update ts-018-whatsapp.md

joelanders:
```
        tcp_blocked = False
        for success, result in results:
            if success == False:
                tcp_blocked = True
```
^ this is what ooniprobe does

### DIFF
--- a/test-specs/ts-018-whatsapp.md
+++ b/test-specs/ts-018-whatsapp.md
@@ -89,7 +89,7 @@ For a given IP address we consider the connection successful if either connectin
 If ALL of the connections succeed then we consider the endpoint to not be
 blocked.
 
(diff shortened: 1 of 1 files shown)


test-specs/ts-018-whatsapp.md
```diff
@@ -89,7 +89,7 @@ For a given IP address we consider the connection successful if either connectin
 If ALL of the connections succeed then we consider the endpoint to not be
 blocked.
 
-Conversely if all attempts fail then we consider the endpoint to be blocked
+Conversely if any attempts fail then we consider the endpoint to be blocked
```
hellais: As we discussed on Slack, this is actually a bug in OONI Probe, not a bug in the spec.

hellais:
Closing.

PR #97, opened 2017-10-29
Ad test helper spec for Parasitic Traceroute

willscott:
In addition to the forward and reverse trace route helpers currently defined, it will be useful to develop a test helper specifically to encode the process of conducting a 'parasitic' trace route on a tcp connection initiated by a client.

This spec addition attempts to formalize the behavior of such a helper.

### DIFF
--- a/test-helpers/th-007-parasitic-traceroute.md
+++ b/test-helpers/th-007-parasitic-traceroute.md
@@ -1,0 +1,66 @@
+# Specification version number
+
+2017-10-29-000
+
+# Specification name
+
+Parasitic Traceroute Test Helper
+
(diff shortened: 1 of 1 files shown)


willscott:
I've completed a second, golang, version of traceroute-as-a-service that will function as a test helper for this form of test: http://github.com/willscott/traas2

hellais:
This should probably include also some mentions of the data format. @willscott do you think you can add to this the specs of the traa2 data format?

I think that is a good start for it, once that is done I think this can be merged.

mScott224 (approved):

mScott224 (approved):

PR #103, opened 2017-12-18
Create ts-022-dpi-fragmentation.md

joelanders:


willscott:
Main issue to flag is that we expect most common clients to not have root access / raw socket support, which will limit the ability of the client to control fragmentation / sending behavior.

A test helper can be privileged and change semantics at this layer in responses back to a client, and learn about the DPI behavior in the Server->Client direction.

bassosimone:
See https://github.com/measurement-kit/measurement-kit/pull/1522 for a possible implementation of this concept. Given that we're moving towards a Go engine, https://github.com/measurement-kit/measurement-kit/pull/1522 will not ever be merged in MK. Therefore, I'm linking it here, so we do remember about its existence, and then I'm going to proceed with closing the PR.

Samboy1990:
DPI-Fragmentation-Spec

mScott224 (approved):

karna2323 (approved):

PR #106, opened 2018-03-29
Added 'that' for clarity.

Tafiti:
line 160: '...but the HTTP request either fails or we get back a HTTP response *that* contains a page we don't expect.'


hellais:
Good call. Thanks!

PR #118, opened 2018-10-31
Add techniques helping to distinguish network outage from filtering.

darkk:
Most of these techniques should be possible to implement within limitations of current Android platform.

Most of these techniques can extent `web_connectivity` as well as other tests.

The techniques are converted from "2018-10 New Nettest" gdoc ver. "dakrk-pr".


techniques/tq-020-0x20-hack-for-possibly-censored-domain.md
```diff
@@ -0,0 +1,10 @@
+# tq-020 0x20-hack for possibly-censored domain
```
darkk: That's unintentional coincidence, I swear! :-)

darkk:
@bassosimone let's coin a term that is slightly better than "censor" and "censorship" (as previously discussed) before merging that.

techniques/tq-001-DNS-UDP-censorship-transparency.md
```diff
@@ -0,0 +1,29 @@
...
+
+DNS testing is fundamental to network measurement to detect censorship as it’s
+a very commonly used technique to implement censorship.
+
+When implementing blocking on a DNS level, censorship equipment will typically
+implement one of two techniques:
+
+- _DNS Hijacking_, whereby a DNS resolver run by the ISP or government will deliver false responses to queries for the users domain. This is usually trivial to circumvent by changing resolver.
+- _DNS Spoofing_, whereby the DNS queries of the user are intercepted and the false response is sent to the user faster than the legitimate response. This tends to be trickier to circumvent.
+
```
bassosimone: My suggestion is to write instead:

> When reasoning about DNS interference, we shall consider the following possibility:
>
>  - _Policy Based DNS interference_, whereby a DNS resolver run by the ISP or government it configured to return specifically altered responses to specific queries. This is usually trivial to circumvent by changing resolver.
>
> - _DNS Spoofing_, whereby there is equipment in the network that listens for DNS queries and sends replies back to the user faster than the legitimate DNS server. This tends to be tricker to circumvent.
>
> - _DNS Transparent Proxy_, whereby all DNS requests sent by a user are routed through a DNS proxy box regardless of the destination DNS server, and the reply is served by the proxy box. This also tends to be trickier to circumvent.

techniques/tq-007-handle-ECMP-while-doing-DNS.md
```diff
@@ -0,0 +1,12 @@
+# tq-007 Handle ECMP while doing DNS
+
+DNS client should `bind()` to random port for every DNS query sent to the same
+server (unless doing traceroutes or other alike measurements) and should not
+reuse same 5-tuple for retries.
```
bassosimone: Here, and perhaps in other places, we should mention using `recvmsg()` and Android to peek at the error queue using `MSG_ERRQUEUE` to extract valuable information where we can.

darkk: I believe, it's same technique having several reasons and applying to all UDP probes we have at our heads right now (DNS, OpenVPN), so I'm renaming it to `tq-007-UDP`.

techniques/tq-001-DNS-UDP-censorship-transparency.md
darkk: Sounds good to me.
I also suggest to write "_Policy Based DNS interference_" as "_Policy Based DNS interference (aka [DNS hijacking](https://en.wikipedia.org/wiki/DNS_hijacking#Manipulation_by_ISPs))_" as this term for this behavior is coined in linked Wikipedia article.
Do you mind?

techniques/tq-007-handle-ECMP-while-doing-DNS.md
bassosimone: Ack!

techniques/tq-001-DNS-UDP-censorship-transparency.md
bassosimone: Not at all, thank you!

techniques/tq-020-0x20-hack-for-possibly-censored-domain.md
bassosimone: lol

techniques/tq-003-DNS-latency-of-non-cached-domain.md
```diff
@@ -0,0 +1,12 @@
...
+It may be tempting to use the latency to resolve
+`random-subdomain.censored.net` as a signal of possible network interference
+touching `censored.net` domain. Indeed, if the recursor hosts a zone for
+`censored.net` it does not have to query the `censored.net` `NS` for the domain. But
+it’s not a 100% clear signal as some [DNSSEC versions may leak cacheable ranges](https://blog.cloudflare.com/dnssec-complexities-and-considerations/)
+of non-existent domains. Also, NSes responsible for the `censored.net` zone may be
+unreachable<sup>[1](#fn1)</sup> and that information may be cached as well, that will also affect
+resolution latency.
+
+<a name="fn1">1</a>: e.g. some israeli authoritative DNS servers were dropping requests from Egypt
```
hellais: It's unclear to me what the technique being highlighted here is. This seems to be mostly about what not to do and not so much about what should be done.

I also don't fully understand what this technique is trying to measure (what "possible network interference").

I would suggest putting a first sentence at the beginning of this explaining what is the intended goal of this.

techniques/tq-007-UDP-information-collection.md
```diff
@@ -0,0 +1,17 @@
+# tq-007 UDP information collection
+
+A UDP client should `bind()` to a random port for every query sent to the same
+server<sup>[1](#fn1)</sup> and should not reuse the same 5-tuple for retries.
+It should enable `IP_RECVERR` and `IP_RECVTTL` to collect more metadata.
+
+By using a random UDP source port:
+- allows us to distinguish ICMP errors delivered via `IP_RECVERR` to `MSG_ERRQUEUE` across queries
+- accounts for ECMP-like<sup>[2](#fn2)</sup> broken paths
+- is BCP to prevent blind DNS/UDP spoofing
```
hellais: What is BCP? Can this acronym be expanded?

techniques/tq-017-request-to-discard-test-helper.md
```diff
@@ -0,0 +1,16 @@
+# tq-017 Request to discard test helper
+
+If a TCP or UDP request triggers an unexpected response it is interesting to send that request to a test helper acting as a discard or dummy server. An unexpected reply highlights possible internet-wide protocol interception. The test helper can do following tricks:
+
+- full packet capture seeking for TCP/IP headers metadata
+- request capture to confirm or refute the mere presence of the request with out of band channel
```
hellais: I edited OOB to be out of band. Is this correct?

I also don’t think I fully understand this point. On which host would the packet capture be gathered and what do you need to refute exactly?

techniques/tq-024-TCP-segmentation.md
```diff
@@ -0,0 +1,19 @@
+# tq-024 TCP segmentation
+
+If a TCP request triggers an unexpected response it is interesting to see if the unexpected response persists if the request is segmented into several distinct packets. There are several segmentation strategies:
+
+- Split the request in the middle of the “badword” (domain, URL, binary fingerprint)
+- Split the request in the beginning of the request to confuse the DPI protocol detection and prevent a reassembly attempt
+- Send the request byte-by-byte with some packet pacing
```
hellais: Did you mean to say “packet spacing”. Can you clarify what this means exactly?

techniques/tq-025-saving-bandwidth-in-repeated-HTTP-requests.md
```diff
@@ -0,0 +1,17 @@
+# tq-025 Saving bandwidth in repeated HTTP requests
+
+HTTP tests have a very high-volume response compared to the request size request to
+censored webpage may have 4 MiB body. If the request is repeated for a hundred times
```
hellais: Did you mean to say instead of “request to censored webpage”, “**response** for a censored webpage”?

I don’t see how a request can weight in at 4MB in body.

techniques/tq-003-DNS-latency-of-non-cached-domain.md
bassosimone: I believe the idea is that you see if there is a rule of `*.censored.net` or not. If there is not, then a query for `${random}.censored.net` will cause recursive resolution. Instead, if there is a policy to always block anything that contains `censored.net` the response MAY arrive earlier because there is no extra latency caused by the recursive resolution. However, there MAY also be caveats as highlighted below.

hellais: Aaah, ok I understand now. Then maybe it would be useful to put that as the first sentence.

techniques/tq-025-saving-bandwidth-in-repeated-HTTP-requests.md
darkk: True. I meant _may lead to 4 MiB body_.
I'll fix & split the sentence in two to decrease confusion.

techniques/tq-024-TCP-segmentation.md
darkk: No, I meant [packet pacing](https://en.wikipedia.org/wiki/TCP_pacing). E.g. sleeping several ms between send() calls to make DPI's life harder and, also, try to avoid some segmentation offload that may kick in (maybe, there should be some explicit "noise" traffic to prevent [LSO/TSO/GSO](https://en.wikipedia.org/wiki/Large_send_offload)... IDK).

techniques/tq-007-UDP-information-collection.md
darkk: That's [Best Current Practice](https://en.wikipedia.org/wiki/Best_current_practice).
Namely, that's [BCP152](https://tools.ietf.org/html/bcp152#section-6.1).
I'll make it a link.

hellais: Edited

techniques/tq-017-request-to-discard-test-helper.md
darkk: I mean packet capture at the test-helper. It may be used to ...
- fingerprint specific TCP/IP header bits if the connection is reset in both directions, so RST packet will have some TTL (that may be same as client's or different), IP ID, maybe it'll have some bad Seq/Ack numbers, maybe TCP Window value will be of interest, and so on
- understand if the server gets the request or not: the request may be dropped by in-path filter, but on-path filter can't do it

OOB is written for following reason: speaking directly to test-helper may be bad idea if IP is blacklisted for a while (e.g. for two hours) after "triggering" DPI device. I've seen that in the wild with attempts to block MTProto proxies, some ISPs apply alike logic to traffic shaping (e.g. Yota's shaper seems to be stateful &mdash; it "learns" the protocol of ServerIP:Port and uses that knowledge to shape connections to same IP:Port using _different_ protocol).

darkk: I believe 1e6be5b resolves the question.

hellais: Thanks for the explaination and adding more details to the text. This is now more clear.

PR #122, opened 2018-12-06
Adds technique to request DNSSEC records

irl-erg:


techniques/tq-xxx-request-dnssec-records.md
```diff
@@ -0,0 +1,37 @@
+tq-xxx Request DNSSEC-related records from resolver
```
bassosimone: I would suggest giving it a number and adding the correct number in the title.

techniques/tq-xxx-request-dnssec-records.md
```diff
@@ -0,0 +1,37 @@
...
+1. Attempt to validate the full chain of DNSSEC signatures back to the root
+   zone for a hostname that is known to have valid DNSSEC signatures set up
+2. If it was possible to retrieve the full chain of signatures, validate them
+
+Implementation issues
+---------------------
+
+This test would require performing a "stub" resolution, where the requests
+that would have been made to the authoritative servers would instead be
+made to the resolver.
```
bassosimone: Why? I mean, wouldn't the packets be intercepted by a middlebox anyway?

irl: This is to emulate a typical client's behavior. The idea is that a typical client will use whatever DNS resolver is given by DHCP, and a DNSSEC validating client will use a stub resolver in this way. It may still be possible to query the servers directly, but this is a different test. I will expand this text to explain both scenarios that could be tested.

bassosimone: Makes sense. Thank you!

irl:
Are there still changes that I should make here? I can add a number for the technique, but I thought it was easier to do that when it is merged to avoid collisions.

bassosimone:
@irl-erg I think this is fine, merging now! Thanks for prodding me :-)

Issue #74, opened 2016-08-08
New major data format version


hellais:
I am going to use this ticket to take note of all the changes that we should make in the next major revision of the data format.

Once we have a sufficiently long list we can make the change and implement them all in ooniprobe, measurement-kit and ooni-pipeline.

So far I have:

* [ ] Change all time stamps to be fully ISO compliant (i.e. include `Z` timezone indicator at the end)

* [ ] Move the `resolver_client` test key as a top level key

hellais:
Related to this: https://github.com/TheTorProject/ooni-probe/issues/689

hellais:
Related tickets:

* https://github.com/ooni/spec/issues/92
* https://github.com/ooni/spec/issues/93
* https://github.com/ooni/spec/issues/107
* https://github.com/ooni/spec/issues/108

bassosimone:
Xref: #113

bassosimone:
My suggestion here is that we proceed as Go 2.0 not as Python 3.0. We have done some work in #160, which seems useful to address some issues. I would suggest we do the same maybe in some months to fix more issues, while preserving back compatibility. We eventually drop old fields.

PR #128, opened 2019-03-13
Write v2.0.0 spec of the collector protocol

bassosimone:
This is backwards compatible with 1.4.0 except that (1) the spec is in a different document; (2) all legacy bits have been explicitly marked as optional. This should hopefully allow us to evolve from the current situation to a situation where we gradually phase out legacy features and stop supporting legacy clients.


bassosimone:
@hellais can you please take a look?

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+# 1.0 System overview
+
+The collector exposes an HTTP API allowing OONI probes to submit the results
+of their measurements. The client and the server SHOULD NOT assume a keep
+alive semantics for the HTTP connections.
+
+A OONI probe that enables `Content-Encoding: gzip` or similar to compress
+request bodies MUST be prepared to receive a `415` response and retry without
+compression. A collector MUST check content-encoding for all requests and,
+if not supported, it MUST send back a `415` response to the client.
```
hellais: This is a very good idea and it would lead to us saving a bunch of bandwidth on the probes.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+
+Measurement submitted to a OONI collector will be archived, processed, and
+published by OONI. How that will happen is out of the scope of this document.
```
hellais: Maybe we can reference our data policy here: https://ooni.io/about/data-policy?

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+
+The collector MUST be exposed as an HTTPS service. It MAY also be exposed as
+a Tor onion service. A [legacy document](
+https://ooni.torproject.org/docs/architecture.html#why-tor-hidden-services)
+explains why the OONI project originally chose to allow for both HTTPS
+and Tor onion service services (henceforth, Onion).
+
+It is also outside of the scope of this section to define the way in which
+a OONI probe discovers the collector API endpoint, as well as how, given
+several endpoints and/or endpoint types (HTTPS, Onion), it choses a specific
```
hellais: typo `s/choses/chooses`

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+the OONI probe and itself. Therefore, a malicious actor won't be able to update
+test results that they have not created.
+
+It is outside of the scope of the collector to provide blocking resistance or
+to conceal to a passive network observer the fact that they are communicating to
+a collector. Such properties are to be provided by other software, e.g. Tor.
+
+Therefore a client implementation of the collector protocol SHOULD allow one
+to specify a [SOCKS5](https://tools.ietf.org/html/rfc1928) proxy where the
+name resolution is performed by the circumvention tool (`socks5h`).
```
hellais: I am unsure about the inclusion of this recommendation (i.e. why we suggest `SHOULD` use SOCKS5 and not for example a VPN or other transport), but I am ok with keeping it.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+When a probe starts a test it will *create* a new report by sending a `POST`
+request conforming to the following, informal specification:
+
+    POST /report
+
+    {
+     "content":
+        (optional, deprecated) `string` it is optionally possible to create a report with
+        already some data inside of it. MUST be a serialized JSON or YAML depending
+        on the value of the 'format' key.
```
hellais: It may be useful to pinpoint the last exact OONI Probe version using this semantic to understand when it's reasonable to stop supporting it.

bassosimone: Right. This information is in `v1.4.0`. I will copy it over.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+        `string` the two-letter country code of the probe as defined in
+        ISO3166-1 alpha-2, or ZZ when undefined (ex. "IT"). This field MUST
+        match this regexp: `^[A-Z]{2}$`.
+
+     "probe_ip":
+        (optional, deprecated) `string` the IP address of the OONI probe client.
+
+     "software_name":
+        `string` the name of the software that is creating a report (ex. "ooni-probe")
+        that MUST match this regexp: `^[0-9A-Za-z_.+-]+$`.
```
hellais: I wonder if these regexps are actually good. It seems redundant to have the `.` (dot) match (which is any character) together with `[0-9A-Za-z]`.
Note: I haven't actually looked at the code to validate if this is what is in the code or not. It may be that we actually wanted to put in here the escaped dot `\.`.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+
+     "software_version":
+        `string` the version of the software creating the report (ex. "0.0.10-beta")
+        that MUST match this regexp: `^[0-9A-Za-z_.+-]+$`.
+
+     "test_name":
+        `string` the name of the test performing the network measurement that
+        MUST match this regexp: `^[a-zA-Z0-9_\\- ]+$`.
```
hellais: Perhaps this regexp also could be aligned with the software_name regexp.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+
+     "test_helper":
+        (optional, deprecated) `string` the name of the required test_helper for this test.
+
+     "test_start_time":
+        (optional) `string` timestamp *in UTC* of when the test was started using
+        the format "%Y-%m-%d %H:%M:%S".
```
hellais: I believe this is an addition to the spec. Since it's new I would suggest we use the ISO standard for ISO timestamp instead of the odd timezone-less OONI timestamp format.
See: https://github.com/ooni/spec/issues/74

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+   
+   - the `test_helper` field is present
+
+   This allows a new implementation to drop support for legacy clients.
+
+4. MUST fail with `5xx` if it cannot generate the report ID (see below) or
+   in case of other failures opening the report.
+
+5. SHOULD exercise care to avoid logging the `probe_ip` field, if set (e.g.
+   by setting it immediately to `null` if such field is not used).
```
hellais: What is the rational behind this?
I am not sure this is within the scope of the spec, especially in light of the fact that `probe_ip` is an optional and deprecated field.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+
+6. if everything is okay, MUST return a `200` response with the body
+   described below.
```
hellais: I believe we currently return a 201 CREATED in case of success for creation.

Perhaps the spec can say that it will return a 2xx status code?

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+        (conditional, deprecated) `string` the address of a test helper that
+        the client requested using the `test_helper_name` field.
+
+      "supported_formats":
+        `list` of strings detailing what are the supported formats for
+        submitted reports. Allowed values are "json" and "yaml".
+    }
+
+New collector implementations MUST&mdash;and existing ones
+SHOULD&mdash;set `Content-Type` using a content type value
```
hellais: Typos? `&mdash;`?

backends/bk-003-collector.md
bassosimone: The dot inside a character set is not special. See the following demo

    ^[0-9A-Za-z_.+-]+$

Regular expression visualization

[Debuggex Demo](https://www.debuggex.com/r/nwnWxuyRSzLx9JAp)

and [a stackoverflow reference](https://stackoverflow.com/questions/19976018).

backends/bk-003-collector.md
bassosimone: It has a space inside because some person submitted measurements with a test_name that contained spaces inside it, IIRC. Did I remember wrong?

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+   a specific implementation is compliant with the format
+   expected by the pipeline (e.g. in `go` with `omitempty`
+   `null` values are removed by a marshal-serialize cycle
+   while the Python parser preserves them).
+
+7. MUST reset the report-specific timer used for automatically
+   closing OPEN reports that have become stale.
+
+8. MAY submit the measurement to some pipeline speed processing
+   layer, but this operation MUST NOT have an impact onto the
```
hellais: I am not 100% sure this is a `MUST NOT`. We may not want to exclude the possibility that the speed processing layer (which should be fast enough to give results to users submitting measurements) could impact the returned status code (ex. a measurement may be deemed unacceptable because the speed layer determined it to be duplicate).

hellais: That said perhaps I am saying this because I don't fully understand the reason to state this.

backends/bk-003-collector.md
bassosimone: It's not an addition. I copied it over from the previous version of the spec where it was also marked as optional. I think we _should not_ change this field because old clients use it. New clients should not provide it, because for JSON reports it's useless: what matters, and what the code should do actually check, are the dates inside of the measurement. This is a relic of when we had the YAML format only and the date was in the initial header.

backends/bk-003-collector.md
bassosimone: The rationale is to be paranoid and make sure that we don't leak PII in the logs by mistake. I think it's good to exercise some PII paranoia when possible. The real impact of this is nearly zero since, as you mention, only very old clients should actually set this field. Yet, I'd like to see the code clear the IP address unconditionally and immediately after parsing.

backends/bk-003-collector.md
hellais: Ah ok my bad. The question still remains of if we should be allowing a dot in software_names and if so why we don't allow it in test_names.

backends/bk-003-collector.md
hellais: Ah ok I see what you mean now. Then perhaps we should mark it as deprecated and check maybe which client versions send this.

backends/bk-003-collector.md
bassosimone: Thanks!

backends/bk-003-collector.md
bassosimone: Yes!

backends/bk-003-collector.md
bassosimone: Thanks!

backends/bk-003-collector.md
bassosimone: So, the reason why here I am mentioning `socks5` is because the actionable action we can take is to start `tor` or `psiphon` and use them to submit. My understanding is that a VPN acts on a different layer, hence I don't believe we'll be able to start it (where we = the OONI app), so it looked like less relevant to include implementation provisions here.

backends/bk-003-collector.md
bassosimone: Good point. I double checked: both `a.collector.ooni.io` and `collector-sandbox.ooni.io` return `200` after the report has been created. TBH, all the `200` can be changed to be `2xx` in theory.

backends/bk-003-collector.md
bassosimone: The fact that the two regexps are not aligned is bad. I think we should relax to the minimum common denominator and add also spaces for test names. (People is anyway gonna use spaces, so we should probably just accept this space oddity.)

backends/bk-003-collector.md
bassosimone: Currently all clients. I didn't notice it was deprecated. I will deprecate and update MK.

backends/bk-003-collector.md
bassosimone: This expands to a "long dash" when visualised in HTML.

backends/bk-003-collector.md
bassosimone: I don't want to keep a connection open waiting for a response. If any middlebox freaks out because the connection is inactive for too much time, then the client doesn't know if the submission succeeded. So, the client may decide to retry and then we have duplicates, and wasted bandwidth.

backends/bk-003-collector.md
bassosimone: @hellais is this answer convincing? How would you include a reference to a VPN here?

backends/bk-003-collector.md
bassosimone: @hellais are you okay with using spaces also here?

backends/bk-003-collector.md
bassosimone: @hellais am I wrong here?

hellais: That may have been the case in some very legacy OONI Probe versions, but we probably want to enforce something a bit more stringent.

FTR I looked up `SELECT DISTINCT test_name FROM software;` on the metadb and I see:
```
base_process_test
base_tcp_test
bridge_reachability
captive_portal
daphn3
dash
dns_consistency
dns_injection
dns_n_http
dns_spoof
facebook_messenger
http_filtering_bypass
http_header_field_manipulation
http_host
http_invalid_request_line
http_keyword_filtering
http_requests
http_test
http_trix
http_url_list
http_vs_https_speed_test
keyword_filtering_detection_based_on_rst_packets
lantern_circumvention_tool_test
meek_fronted_requests
meek_fronted_requests_test
multi_ndt
multi_protocol_traceroute
ndt
netalyzr_wrapper
net_test
ooni_test
openvpn_client_test
p2p_bittorrent_test
parasitic_traceroute
peer_locator
psiphon_circumvention_tool_test
psiphon_test
sip_requests_test
spam
squid_test
ssl_observatory
tcp_connect
telegram
test_access_to_tor_directory_authorities
this_test_is_nameless
tls_handshake
tor_bandwidth_scan
tor_exit_ip_test
tor_http_requests_test
tor_ssl_observatory
vanilla_tor
web_connectivity
website_probe
whatsapp
```

backends/bk-003-collector.md
hellais: Yes this makes sense. Let's keep this here since it's in any case just a recommendation.

backends/bk-003-collector.md
hellais: See my comment on test_names. I think we can probably also drop the space.

Here is FTR all the existing software_name in the DB:
```
alladin_Nettest
measurement_kit
ooniprobe
ooniprobe-android
ooniprobe-desktop
ooniprobe_ios
ooniprobe-ios
spam
```

backends/bk-003-collector.md
hellais: Sorry I misunderstood your comment. I checked `ooni-backend` and it seems like the current collector relies on this value to generate a valid filename: https://github.com/TheTorProject/ooni-backend/blob/master/oonib/report/handlers.py#L117.

It may however be a good idea to deprecate it as we can derive it from other fields anyways.

backends/bk-003-collector.md
hellais: Ok then in that case let's define it as 200.

backends/bk-003-collector.md
hellais: Ah I see. Can we replace it with the UTF8 character or just a simple dash to make it more readable in plaintext?

backends/bk-003-collector.md
bassosimone: Ok, so no need to use a space.

backends/bk-003-collector.md
bassosimone: Thanks! No need to use a space, then.

backends/bk-003-collector.md
hellais: Ok I understand the concern. Can we then perhaps phrase this in terms of what performance requirements to deliver a response to the client we expect the collector to have as opposed to excluding some system design a-priori?
What I mean is that we may be able to achieve sub-second latency even when submitting to the speed layer, conversely we may have latency higher than one second even if we don't submit to a speed layer.

backends/bk-003-collector.md
bassosimone: So, here I can actually see two approaches:

1. the new collector generates the file name using its current time, so we deprecate the field

2. we make the field non-optional and we use it to generate the file name

The latter approach is the one of the existing collector. (Where, BTW, things work just because all existing implementations always include that field.)

I think actually 1. is more robust since this field is really a relic IMHO.

backends/bk-003-collector.md
bassosimone: Sure

backends/bk-003-collector.md
hellais: Yeah I agree. Moreover even in the current collector it's optional and when it's missing the collector use the approach from 1:
https://github.com/TheTorProject/ooni-backend/blob/master/oonib/report/handlers.py#L118

Hence I think it does not make sense to keep this extra logic lying around if we can avoid it.

backends/bk-003-collector.md
hellais: Understood. I think it's reasonable to keep it in then.

backends/bk-003-collector.md
bassosimone: > Can we then perhaps phrase this in terms of what performance requirements to deliver a response to the client we expect the collector to have as opposed to excluding some system design a-priori?

Sure

backends/bk-003-collector.md
darkk: I'm not 100% confident, but I expect `software.test_name` to store the test_name [AFTER normalisation](https://github.com/ooni/pipeline/blob/83bafc99f7e3b19d6cb72e79cbf191e5566969e7/af/shovel/daily_workflow.py#L23-L91).

So I'm unsure about the SQL query results relevance for the discussion about collector.

hellais: Ah OK I see. I had forgotten about that.

According to the code you linked it the normalisations that happen are:
1. Checks if the test_name is inside of the map `test_name_mapping` (https://github.com/ooni/pipeline/blob/83bafc99f7e3b19d6cb72e79cbf191e5566969e7/af/shovel/daily_workflow.py#L23)
2. If it's in the map it replaces it with the value in there if not it just lower cases it

I believe those transformations are applied due to some very old OONI Probes submitting results with spaces in them (ex. `HTTP Invalid Request Line`), but I suspect this was an artifact from pre-1.0 OONI Probe.

It's probably still wise to tighten up the regexp and not allow spaces in test_names as I don't expect modern OONI Probes to have spaces in them anymore.

You are though right to point our that in order to validate this we should probably look at pre-autoclaving measurements.

backends/bk-003-collector.md
darkk: Does OONI Probe submit real `probe_ip` to the collector in the case when the `probe_ip` is not added as a field to the measurement?

If it does -- it's the significant issue of the probe. And, yes, collector needs workaround, but link to relevant ticket or the commit should be probably added for the reference.

If the `probe_ip` is sent only after user's opting into IP disclosure &mdash; I don't quite understand why collector should be worried about it. Should the collector be, for example, worried about `X-Forwarded-For` HTTP header as well?

hellais: I believe this was a feature that was specified in the backend, but as far as I know was never implemented in any real probes. The rational behind this was that it would allow collectors behind onion services to do some geoip resolution on the collector side, but it never happened.

backends/bk-003-collector.md
```diff
@@ -0,0 +1,519 @@
...
+Modern OONI probes represent a set of logically-related measurements
+(a *report*) as a set of separate JSON documents each containing a
+JSON object. Legacy OONI probes represent a report as a single YAML
+document containing a header and several YAML objects. Users still using
+the old YAML based format are encouraged to upgrade to JSON ASAP. A
+server side implementation of the collector MUST support the JSON format.
+
+The collector MUST be exposed as an HTTPS service. It MAY also be exposed as
+a Tor onion service. A [legacy document](
+https://ooni.torproject.org/docs/architecture.html#why-tor-hidden-services)
```
bassosimone: Apparently the anchor is gone in the referred document. Reported by @darkk.

backends/bk-003-collector.md
bassosimone: @darkk FWIW, MK always sets `probe_ip` to `null`. I think I'll actually change the code to avoid sending it.

backends/bk-003-collector.md
darkk: > in the DB

FTR, there ARE some reports those are not properly reflected in the database due to some failures here and there.

backends/bk-003-collector.md
bassosimone: Correct, it is not. The previous version of this document mentions OONI backend not OONI probe. I guess we should just setup metrics to scrape and eventually pull the plug for some fields.

backends/bk-003-collector.md
bassosimone: So, it seems that [very old code accepted a space since the beginning](https://github.com/TheTorProject/ooni-backend/commit/5b4cac47852e685ffabc7bbededf9ae08fe535af#diff-3ed66cda14ffcaa8975e48a95e1f7df7R16). The current code still contains and allows for space. Yet, I am in favour of dropping the space. I am going to recommend in the spec to create alerts for failing submissions, so we will know.

backends/bk-003-collector.md
bassosimone: Same comment as above. OK for dropping space. Requirement for alerts.

backends/bk-003-collector.md
bassosimone: Snap, I forgot about this. Will do in another commit.

PR #135, opened 2019-04-30
collector: clarify software_{name,version} meaning

### LINKED ISSUE #134: collect: ambiguous software name and version
When opening a report with the collector, the spec should clarify that the software name and software version are the ones of the code that is submitting a report, not the ones of the code that created the report.

The specific problem here is that “create” is ambiguous because it could either mean the act of submitting a report (where the first operation is called “create”) and the act of creating a file report on disk.

bassosimone:
Closes #134


hellais:
lgtm

PR #153, opened 2019-08-13
collector.md: clarify response body expectations

bassosimone:


hellais:
Are we sure we want to be dropping the error key? I think having an error key where we communicate to the client/user why the report creation failed is useful especially if the user is writing a new OONI client and they need to figure out why their reports are not getting accepted.

bassosimone:
> Are we sure we want to be dropping the error key? I think having an error key where we communicate to the client/user why the report creation failed is useful especially if the user is writing a new OONI client and they need to figure out why their reports are not getting accepted.

In general I am against providing too much hints to clients when there is a failure and prefer to err on the side of just saying there was a failure. In any case, the wording does not prevent you to write a collector that provides such information at all. With this wording the main change is that we are not _promising_ to client implementors that there will be an _actionable_ key with that name that can used for specific purposes. The main reasons why I changed the wording to say that there will always be a body are (1) to allow for uniform parsing and (2) to support use cases like the one you described. (Even though in my mind that was temporary when developing rather than being something baked into the implementation of the collector.)

hellais:
How about specifying that there CAN be an error key, but not making it a requirement?

bassosimone:
> How about specifying that there CAN be an error key, but not making it a requirement?

What @darkk suggested when we did a previous round of review of these specification [*] was that we would focus on what MUST happen, since we're few and there's already complexity, so there is generally little point in spelling out what MAY happen. (That is, what is not forbidden is allowed.)

[*] I initially balked at this suggestion, but later I managed to embrace it, and now I believe it is a reasonable approach for programming in the small.

hellais:
> we should focus on what MUST happen, since we're few and there's already complexity, so there is generally little point in spelling out what MAY happen. (That is, what is not forbidden is allowed.)

Yes this does seem to be a good principle in general, however I was highlighting this as we currently do return the `error` key inside of the existing implementation of the collector (and also the new one) and hence I find this spec to be incomplete with the lack of this key.

That said I guess a developer will probably notice the key nonetheless and think to use it, so I am also alright with not including it, though I still believe it's a small effort to add it and maintain these extra words to provide less surprises to users of these backends.

bassosimone:
> [...] I still believe it's a small effort to add it and maintain these extra words to provide less surprises to users of these backends.

OK, I can probably have some wording that could _save the goat and the cabbage_

bassosimone:
@hellais PTAL?

hellais:
Yeah this looks good.

PR #156, opened 2019-09-28
Update tq-027-stuffed-TLS-ClientHello.md

ValdikSS:


techniques/tq-027-stuffed-TLS-ClientHello.md
```diff
@@ -5,9 +5,11 @@ If a TLS request triggers an unexpected response it is interesting to try
 different behavior. Possible fields for "stuffing" are:
 
 - Session ID, 256 bytes
-- Cipher Suites, 65536 bytes
+- Cipher Suites, ~16 Kbytes
 - Compression Methods, 256 bytes
-- Extensions, 65536 bytes
+- Extensions (especially Padding), ~16 Kbytes
+
+The whole ClientHello packet is limited to 16 Kbytes.
```
darkk: Does the limit come from [RFC5246 6.2.1](https://tools.ietf.org/html/rfc5246#section-6.2.1) or from some other source? Have you, maybe, tried to abuse TLS fragmentation to get some interesting results with even larger ClientHello?

ValdikSS: Yes, also https://tools.ietf.org/html/rfc8449

>   Receiving large protected records can be particularly difficult for a
   device with limited operating memory.  TLS versions 1.2 [RFC5246] and
   earlier permit senders to generate records 16384 octets in size, plus
   any expansion from compression and protection up to 2048 octets
   (though typically this expansion is only 16 octets).  TLS 1.3 reduces
   the allowance for expansion to 256 octets.  Allocating up to 18K of
   memory for ciphertext is beyond the capacity of some implementations.

I haven't tried TLS fragments, as far as I understand, it should be supported on the both sides.

darkk: My current understanding is that TLS fragments must be supported by a "valid" TLS protocol client/server. [RFC5246 D.4](https://tools.ietf.org/html/rfc5246#appendix-D.4) has a really nice paragraph:

> Do you correctly handle handshake messages that are fragmented to multiple TLS records (see Section 6.2.1)? Including corner cases like **a ClientHello that is split to several small fragments**? Do you fragment **handshake messages that exceed the maximum fragment       size**?

But I've not actually read the TLS1.2 spec as a whole, so I may misinterpret those words.

bassosimone:
I am going to merge this. Thanks @ValdikSS and @darkk!

PR #159, opened 2019-10-08
Write spec for SNI blocking experiment

bassosimone:


nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+
+- "interesting" if there is no failure when using the random SNI
+
+- "normal" if there is no failure with proper SNI and there is an error
+indicating that the certificate does not apply to the random SNI
+
+- "suspicious" if there is any failure with proper SNI and there is an error
+indicating that the certificate does not apply to the random SNI
+
+- "🤷" otherwise
```
hellais: I suggest using standard ascii characters for fields that are expected to be parsed.

Maybe this can be "unknown"?

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+performing a TLS handshake to the input endpoint using proper SNI. This key
+will be null in case there is no error.
+
+The `test_keys.failure_with_random_sni` contains the error of connecting and
+performing a TLS handshake to the input endpoint using a random SNI. This
+key will be null in case there is no error.
+
+The `test_keys.behavior` key is determined as follows:
+
+- "interesting" if there is no failure when using the random SNI
```
hellais: Is this even possible? In which circumstances can this be the case?

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+
+- "normal" if there is no failure with proper SNI and there is an error
+indicating that the certificate does not apply to the random SNI
+
+- "suspicious" if there is any failure with proper SNI and there is an error
+indicating that the certificate does not apply to the random SNI
```
hellais: It's worth mentioning that this may also be due to the website being unavailable due to other reasons (for example the website is down).

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+
+0.1.0
+
+# Specification name
+
+SNI blocking (`sni_blocking`)
+
+# Test preconditions
+
+  * An internet connection
```
hellais: I think where this test is really useful is in cases in which you already have a site which you suspect to be blocked by means of SNI filtering and you want to confirm that is in fact the case. I think this should be documented somewhere in the specification.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+
+are all valid entries.
+
+# Test description
+
+For every input endpoint we:
+
+1. connect and perform a TLS handshake
+
+2. repeat forcing a random SNI
```
hellais: I think it would also be interesting to repeat the experiment with variations of the domain to understand how the SNI fingerprinting/filtering is happening.

Variations that can be done include (assuming you are measuring `example.com`):
* `[random-prefix]example.com`
* `[random-prefix].example.com`
* `example.com[random-suffix]`
* `[random-prefix]example.com[random-suffix]`

Of these I think the one for the subdomain is the most important one as it allows us to learn if the block is affecting every subdomain of the site.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,114 @@
...
+
+The `test_keys.failure_with_proper_sni` contains the error of connecting and
+performing a TLS handshake to the input endpoint using proper SNI. This key
+will be null in case there is no error.
+
+The `test_keys.failure_with_random_sni` contains the error of connecting and
+performing a TLS handshake to the input endpoint using a random SNI. This
+key will be null in case there is no error.
+
+The `test_keys.behavior` key is determined as follows:
```
hellais: In thinking more about this I wonder if in fact we want to have this logic live inside of the probe and if we shouldn't instead leave the determination of the results of this experiment to who analyses the data.

These `behavior` keys appear to just be a very simple logic operation on the `failure_with_proper_sni` & `failure_with_random_sni` keys which can be done by any consumer of the data without necessarily relying on the probe side computation.

Maybe we can still document these failure cases in the specification, explaining more in detail why one is an interesting case, why another is normal and why another is suspicious, but not generating it in the data format itself as it's a duplcation of data.

Moreover if we decide to enrich this test with more checks (like the ones I proposed in: https://github.com/ooni/spec/pull/159/files#r332989964) we would have to update the semantics of the derived key making the processing logic more tricky.

nettests/ts-024-sni-blocking.md
bassosimone: If there is a certificate installed that allows a proxy to sign on the fly every request for every domain for the purpose of decrypting traffic (e.g. Kazahstan)?

nettests/ts-024-sni-blocking.md
bassosimone: So, the website being down results in a different error in the second request. What could happen is the website transiently being down in the first request.

bassosimone: I tried to capture this concept with "suspicious" but you're right that I can be more explicit.

nettests/ts-024-sni-blocking.md
bassosimone: Agreed. I want to keep the first version of the spec lean, also to implement, and thus I have documented the lack of the above functionality as "Limitations" of this current spec.

nettests/ts-024-sni-blocking.md
bassosimone: This is a good suggestion. I was also asking myself the same question. So, I will remove that part of the spec and just explain what we expect to see.

nettests/ts-024-sni-blocking.md
hellais: > If there is a certificate installed that allows a proxy to sign on the fly every request for every domain for the purpose of decrypting traffic

I would be very surprised if that were the case. AFAIK the certificate is generated in advance for the specific target they plan to perform a MITM on. Doing it on the fly would mean a fair amount of computing resources and I don't think is really feasible.

bassosimone:
@hellais I have addressed your feedback, I hope! Please, take another look!

nettests/ts-024-sni-blocking.md
bassosimone: We don't need to mention this in the specification then. My basic point is just that I'd be interested such a case, and that it would surprise me to see this data point.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,119 @@
...
+
+SNI blocking (`sni_blocking`)
+
+# Test preconditions
+
+  * An internet connection
+
+# Expected impact
+
+Understanding whether there is SNI based blocking. This experiment is useful
```
fortuna: Minor thing: a more precise term I'm tending towards is "SNI-triggered blocking". It's clearer about the fact that SNI is used for detection and that the blocking action can vary.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Expected output
+
+The key `test_keys` key object will be added to the report.
+
+The `test_keys.failure_with_proper_sni` contains the error of connecting and
+performing a TLS handshake to the input endpoint using proper SNI. This key
+will be null in case there is no error.
+
+The `test_keys.failure_with_random_sni` contains the error of connecting and
```
fortuna: It seems this is a string. However, the type of error matters.
Forcing the test data user process the string is counter-productive. We should do some weightlifting for them instead.

To start with, it would help to have a structured format. A numerical or enum error code of sorts + a text message. This way users can use the error codes to interpret the results. We'll need to define those error codes.

We also need to describe how to conclude blocking from the results. The main output should be: sni_test: BLOCKED | NOT_BLOCKED | INCONCLUSIVE and have an explanation for the choice.

You should start with the connection with proper SNI:
- If you fail the TCP connection, then you can't say anything about SNI, but you're likely blocked by IP. You can have: INCONCLUSIVE: failed to establish TCP connection. And add whatever underlying error (reset, timeout, ...)
- If you succeed TCP, proceed
- Try the TLS handshake. If the connection gets closed or times out without response, that's likely blocking. Run the test with random sni
- If you get the ServerHello, but it fails to validate, wither the server is misconfigured, or there's MITM.
- If the ServerHello validates, you can say NOT_BLOCKED: successful TLS handshake.

We can keep developing this. My point is, the current two fields will burden the user and not give all the information we can. We need to give one test output, with a clear explanation that justifies the result.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,119 @@
...
+www.facebook.com:443
+```
+
+are all valid entries.
+
+# Test description
+
+For every input endpoint we:
+
+1. connect and perform a TLS handshake
```
fortuna: What IP address are you connecting to?
How will we differentiate SNI blocking from DNS blocking?

We need an SNI test that is not vulnerable to DNS blocking. Some alternatives:
1. Get the IP address from a secure OONI backend
2. Use an IP address for a TCP server that you know is not blocked.

I feel like it would be more productive to create a new web_connectivity test instead of a SNI-specific one, so we can combine conclusions from different tests more easily. It will make the user's like a lot easier.

bassosimone: This and https://github.com/ooni/spec/pull/159#discussion_r333082323 are great questions, thank you! My initial answer regarding the structure of errors and how to differentiate DNS blocking from SNI blocking is that I was planning (in a follow-up amendment to the initial spec) to add `netx` events to the report.

This is an example of how such events look like when using the default resolver.

First we have the name lookup:

```JSON
{"Resolve": {
  "Addresses": ["104.198.14.52"],
  "Duration": 76249751,
  "Error": null,
  "Hostname":"ooni.io",
  "Time":81148419}}
```

so, the name resolution did look good. Then we connect:

```JSON
{"Connect": {
  "Duration": 179552191,
  "Error": null,
  "Network": "tcp",
  "RemoteAddress":  "104.198.14.52:443",
  "Time":182450044}}
```

so, we know that the TCP connect was also good. Then TLS handshake:

```JSON
{"Write":{"Duration":73575,"Error":null,"NumBytes":283,"Time":183650757}}
```

This is a write during the TLS handshake and we have more writes and reads that I'm glossing over to arrive quickly to the interesting message:

```JSON
{
  "TLSHandshake": {
    "Config": {
      "NextProtos": [
        "h2",
        "http/1.1"
      ],
      "ServerName": "fhggh.cih"
    },
    "ConnectionState": {
      "CipherSuite": 0,
      "NegotiatedProtocol": "",
      "NegotiatedProtocolIsMutual": false,
      "PeerCertificates": null,
      "Version": 0
    },
    "ConnID": 1,
    "Duration": 448853296,
    "Error": {
      "Certificate": {
        "Raw": "MIIGIzCCBQugAwIBAgIQC1W/C9syOFclqIEum..."
      },
      "Host": "fhggh.cih"
    }
}
```

This tells us that the phase that failed was the TLS handshake. The specific error in this case is `x509.HostnameError`. I have so far resisted the temptation to wrap errors in a way that gives them a name, but it would probably beneficial in this case to know that this error is indeed a `x509.HostnameError`. (The string you commented about is the serialisation of this error and it seems errors in Go produce different values when serialised as JSON and when serialised using the `.Error()` method.)

Back to my original point: the addition of this stream of `netx` events to the report is, in my view, quite effective in telling us what step has gone wrong. The raw stream is missing the information of what is the type of the error (and I guess Go 1.13 `Unwrap` could help).

(Unrelated to this specific conversation, my plan for all OONI tests is to include in all of them these events that I have showed here, so one can understand what specific operation went wrong in which specific phase of an experiment; i.e. resolve, connect, handshake, etc.)

Looking forward to have a more detailed chat on this soon!

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,436 @@
...
+may possibly be useful to detect these specific cases.
+
+For this reason, it is ideal to select as test helper endpoint one that
+knows how to handle one the control SNI, which is what the implementation
+of this experiment should be doing by default.
+
+## Example output sample
+
+```JSON
+{
```
fortuna: It would be great to see the conclusion of the test in this output. Whether it's blocked, not blocked or inconclusive. And why. The `failure` field is not that.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,436 @@
...
+input is a URL, the experiment will extract the domain from the URL and
+use that as target SNI, ignoring any scheme, port, path, etc.
+
+The user should be able to specify the above parameters from the CLI.
+
+# Test description
+
+For every `target`, this experiment will:
+
+1. if `target` is a URL, parse it and set `target` as the URL's hostname
```
fortuna: This might work better:
1. Establish a TCP connection to the `testhelper`. In case of failure (connection refused, timeout), abort and return `INCONCLUSIVE - can't connect to testhelper`. The `testhelper` may be blocked or down, we should track that.
1. Send ClientHello using `target` as SNI.
1. If you get a TCP RST or FIN after the ClientHello, return `BLOCKED`
1. If you get a ServerHello with a certificate for the `target` domain:
    - If it has a valid CA chain, then the TLS is successful and you can return `NOT BLOCKED`. This will only happen if the test site is co-hosted with the testhelper (e.g. they are both on Cloudflare)
    - If the certificate does not have a valid CA chain, then return `MITM`
1. If you get a ServerHello for a different domain, return `NOT BLOCKED`.  This assumes that the `testHelper` always returns a ServerHello for a domain it can serve (as opposed to closing the connection, for example).
1. If it times out, you can try again, or you can try asking for the `testhelper` domain. If it's successful, then it's `LIKELY_BLOCKED` by dropping packets. Otherwise, the `testhelper` is broken and the result `INCONCLUSIVE`.

Notice that there's no need for a control, except in the time out case (to verify the testHelper is not broken). But the testHelper can be tested once before a series of tests.

It may be worth to emphasize that the reason for having the `testhelper` is to be able to test SNI-triggered blocking in the presence of DNS or IP censorship. Otherwise we could just talk to the actual server instead.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,436 @@
...
+  },
+  "test_name": "sni_blocking",
+  "test_start_time": "2020-01-28 15:27:18",
+  "test_version": "0.0.1"
+}
+```
+
+# Privacy considerations                                     
+
+This nettest may be less intrusive than other nettests that measure
```
fortuna: Two benefits are that we are not issuing a DNS query for the sensitive domain and we are not connecting to the sensitive IP. It's easier to show that the user was not trying to access whatever site may be prohibited.

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,436 @@
...
+blocking of a specific host by connecting directly to it.
+                                                                
+# Packet capture considerations                      
+
+This test does not capture packets by default.
+
+# Acknowledgments
+
+In Autumn 2019, [@fortuna](https://github.com/fortuna) proposed, designed,
```
fortuna: Thanks! ☺️

nettests/ts-024-sni-blocking.md
```diff
@@ -0,0 +1,436 @@
...
+
+- `queries` (`[]Query`; nullable): see `df-002-dnst`;
+
+- `requests` (`[]Transaction`; nullable): see `df-001-httpt`;
+
+- `sni` (`string`): SNI being used;
+
+- `tcp_connect` (`[]TCPConnect`; nullable): see `df-005-tcpconnect`;
+
+- `th_address` (`string`): address of the test helper;
```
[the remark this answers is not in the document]
bassosimone: I'll add a `(see above)` to that sentence; above it says it may also be a domain name. I've tagged you into a discussion oriented issue where we want to figure out a good test helper strategy.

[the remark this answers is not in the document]
bassosimone: https://github.com/ooni/spec/commit/45741cce895e3ab380569b3dc9d08a9ab7e1e206

nettests/ts-024-sni-blocking.md
bassosimone: Yeah

nettests/ts-024-sni-blocking.md
bassosimone: Thanks, I will take note about reordering the operation as you suggest and about using this logic to set the toplevel keys you mentioned above. I have have this comment/question:

> [...] This will only happen if the test site is co-hosted with the testhelper (e.g. they are both on Cloudflare) [...]

I believe this also happens if there is surveillance with a transparent proxy in the middle that logs the SNI that you are connecting to. This event may be unlikely, but maybe it's better to flag this differently so that later we can check whether it was Cloudflare, or something else, and whether we believe that IP was legitimately serving that domain. Also, this reasoning leads me to think that perhaps we want to annotate the returned IP address with its ASN.

nettests/ts-024-sni-blocking.md
bassosimone: See https://github.com/ooni/probe-engine/issues/309

nettests/ts-024-sni-blocking.md
bassosimone: See https://github.com/ooni/probe-engine/issues/309

nettests/ts-024-sni-blocking.md
bassosimone: See https://github.com/ooni/probe-engine/issues/309

nettests/ts-024-sni-blocking.md
darkk: > I would be very surprised if that were the case

FYI, certificates were generated and signed on the fly in KZ case. That's how [a rogue certificate for `online.sberbank.kz` name](https://gist.github.com/darkk/8356d84eabeb472120a92b7e10626cf4#file-02-ripe-atlas-sberbank-certificate-pem) appeared :-)
But, yes, the generated certificates were cached to save some CPU.

hellais: Oh that's cool, thanks for the reference on that @darkk !

PR #160, opened 2019-10-11
Bump data format version to v0.2.1

bassosimone:
1. document `resolver_ip` as a top-level key.

2. take some time to go through really collected measurements and
make sure the specification is in sync with reality, and specifically
document historical fields and their canonical serialization when
they are not set. Should help the Go collector work.

3. rewrite the specification in a format that is more agile to edit
in the future, without changing its meaning.

4. since too many changes were already in, just apply the minimal
set of changes to document the new `headers_overflow` field.


data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
-    "test_keys": {
-        "The keys that are specific to the test"
-    }
+where `{timestamp}` is like `20151122T103202Z`.
+
+- `report_id` (`string`): identifier of a set of related measurements
+generated by OONI backends when submitting one or more measurements.
+
+- `resolver_ip` (`string`; since 0.2.1): IP of the DNS resolver used by
+the probe, as determined by the measurement engine.
```
hellais: We should maybe document that some version of OONI Probe put this key inside of the `test_keys` and if it's easy mention explicitly which are those versions.

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
-contain instead all the keys that are specific to the test in question.
+contain instead all the keys that are specific to the test in question, as well
+as all the keys generated by the test template.
+
+Keys starting with `x-` are always permitted anywhere. They are
+experimental and should not be relied upon in the long term.
+
+As an historical note, the `test_keys.resolver_ip` key has been set by
+Measurement Kit implementations inside of the `test_keys` even though it
+should have been part of the top-level keys. This has now been fixed.
```
hellais: Ah ok I see this is mentioned in here.

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
 
 This specification is of the basic data format common to all ooniprobe test
 outputs.
 
 Every entry contains the following common fields. The `test_keys` key will
-contain instead all the keys that are specific to the test in question.
+contain instead all the keys that are specific to the test in question, as well
+as all the keys generated by the test template.
+
+Keys starting with `x-` are always permitted anywhere. They are
```
hellais: Maybe this should be `x_` for consistency with the `snake_case` of other keys?

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
 
 ## Specification
 
-```
-{
-    "input": "If the test takes an input this will contain the value of it"
-        " these can be for example URLs, hostnames, IPs, etc.",
+The following fields are defined:
+
+- `annotations` (`map[string]scalar`; optional): key-value annotations to the
```
hellais: Shouldn't this be `map[string]string`?

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
-    "input_hashes": "A list of the SHA256 hash of encoded "
-        "as hex of the inputs to this test.",
+  - `network_type` (`string`): one of:
 
-    "id": "This is an identifier of this particular measurement",
+    - mobile
 
-    "bucket_date": "A date in the format of %Y-%m-%d that indicates "
-        "when the report was processed by the data pipeline"
+    - wifi
```
hellais: We should explain the meaning of `wifi` and `mobile` in the context of the mobile app. We probably also want to document what should be done in the case of a desktop probe.

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
-        "UTC time coordinates (ex. 2015-08-24 12:02:23) (Note: ooniprobe <= 1.4.0 generates skewed timing information)",
+- `input` (`string`; optional): if this experiment accepts any input, the
+input that was used to produce this measurement. For example, the Web
+Connectivity experiment uses URLs as input.
 
-    "test_start_time": "Timestamp of when the test was started in "
-        "UTC time coordinates (ex. 2015-08-24 12:02:23) (Note: ooniprobe <= 1.4.0 generates skewed timing information)",
+- `input_hashes` (`[]string`; optional; deprecated): historical field that
+used to contain the SHA256s of all inputs provided to the experiment. Modern
+implementations typically emit an empty list.
```
hellais: Maybe we can make it so modern implementations don't include this field at all?

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
 
+- `measurement_start_time` (`string`): time when this measurement was
+started in UTC, using the `"2006-01-02 08:04:05"` format. Note that
+ooniprobe <= 1.4.0 generates skewed time information.
+
+- `options` (`[]string`; optional): list of options passed on the
+command line when running this specific experiment. Modern implementations
+typically emit an empty list.
```
hellais: Do we want to mark this as deprecated as well and make modern clients drop it entirely?

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
+
+- `probe_asn` (`string`): AS Number of the probe (prefixed by AS, e.g.,
+`"AS1234"`), or `"AS0"` if the user does not want to share their ASN.
+
+- `probe_cc` (`string`): two letter country code of the probe (e.g.,
+`"IT"`) or `"ZZ"` if the user does not want to share their country code.
+
+- `probe_city` (`string`; optional): name of the city where the
+measurement was run. If the user does not want to share this information,
+this field should be set to `null`.
```
hellais: I believe no modern client is setting this field anymore, maybe we should deprecate it?

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
 
-    "test_name": "The name of the test that generated "
-        "this measurement (ex. http_requests)",
+    - add `resolver_ip` as a top-level key.
 
-    "test_version": "",
+    - introduce `headers_overflow` in the HTTP base data format.
 
-    "test_runtime": null,
+- `id` (`string`): UUID4 identifying this specific measurement.
```
hellais: I think we should mention that this ID is generated by the client and as such should not be trusted as a unique ID for the measurement.

I also think we should consider dropping this as it's unclear the usefulness of it.

data-formats/df-000-base.md
```diff
@@ -10,135 +10,225 @@ The output of test is the composition of:
...
+used, e.g., by Web Connectivity. The former is typically used when there
+can only be a single type of backend. The latter when more types are possible.
+
+- `test_keys` (`object`): object containing specific keys that depend
+upon the specific network experiment that we're running.
+
+- `test_name` (`string`): name of the experiment in snake case. For example,
+Web Connectivity SHOULD be indicated as `"web_connectivity"`.
+
+- `test_runtime` (`float`): runtime of this specific measurement in seconds.
```
hellais: Can we be more specific as to what is included in the `test_runtime`? I believe that we currently include as part of the test runtime also the time it takes to perform the measurement upload, is that the case?
We should also perhaps document that it includes the time it takes to communicate with any test helper if they are used.

data-formats/df-000-base.md
bassosimone: It is easy for Desktop (since `3.0.0`). We need to remember to update this for mobile, for which we don't have a roadmap for switching to `ooni/probe-engine` yet.

data-formats/df-000-base.md
bassosimone: Yeah, I can make it a bit more specific anyway

data-formats/df-000-base.md
bassosimone: Yes, good point!

data-formats/df-000-base.md
bassosimone: I am no sure. I initially wrote that and then wondered if it was correct. How do we know?

data-formats/df-001-httpt.md
```diff
@@ -27,7 +27,14 @@ data format.
                 "will be in here, otherwise it is set to null",
 
             "headers": {
-                "Header-Name": "Header-Value"
+                "Header-Name": "first value of header with such name"
+            },
+
+            "headers_overflow": {
```
hellais: Why don't we just create a new key which encodes the headers in a robust way which isn't affected by the issues of the key-value scheme.

I say this because from the perspective of a consumer of the data it's easier for me to just parse one single field rather than have to consider two different fields and merge the two.

This way we can keep the old format for legacy purposes, but also start moving to the newer format with new processing logics.

Moreover I think that having `headers_overflow` be again a dictionary we aren't addressing one of the issues with the current format which is of preserving the order in which headers are received which may be important from the perspective of fingerprinting middleboxes.

My suggestion is to use a list to encode the headers, something like:

```
"headers_list": [
{"name": "My-Fancy-Header", "value": "My Header Value"},
{"name": "Host", "value": "apple.com"},
{"name": "X-Antani", "value": "Prematurata"},
{"name": "X-Antani", "value": "Supercazzola"},
]
```

data-formats/df-000-base.md
bassosimone: Ok for explaining. I believe what to write for Desktop I cannot fill it yet, since I dunno.

data-formats/df-000-base.md
bassosimone: Sure, even better!

data-formats/df-000-base.md
bassosimone: Yes!

data-formats/df-000-base.md
bassosimone: Yes!

data-formats/df-000-base.md
bassosimone: I need to read the code to be sure :-). Will do.

data-formats/df-001-httpt.md
bassosimone: Sure, I can do that. I was conservative to avoid duplicating data.

data-formats/df-000-base.md
hellais: AFAIK we haven't used anything other than string values in them. For example the ooniprobe 2.x CLI supported passing annotations as key-values that were encoded as strings. I am not aware of other ways to encode annotations.

That said since we are writing the spec now we can decide if we want to be more strict or more lax in this.

I think it's probably wise to fix it to `map[string]string`, but mention that we aren't fully sure if legacy clients are going to be fully compliant and the consumers of this data should be robust for this.

data-formats/df-000-base.md
bassosimone: The usefulness that I see is to refer to another measurement and say that this measurement is a follow-up to that measurement. We can trust a client to give names to their measurements, I guess. If we don't, then what schema can we apply?

hellais: >  this measurement is a follow-up to that measurement

I think that to encode this type of information in data that is uploaded to collectors we should use an ID which is generated by OONI backend and therefore can be trusted.

We don't want to give the ability to attach a measurement to existing ids which are created by the client.

I suppose it can be useful as an identifier as long as the measurement sits on the users device, but once it goes into the collector and is published I think this should stop being trusted as a good source of unique measurement identification.

data-formats/df-000-base.md
bassosimone: okay, let's do it like this

data-formats/df-000-base.md
bassosimone: > I suppose it can be useful as an identifier as long as the measurement sits on the users device, but once it goes into the collector and is published I think this should stop being trusted as a good source of unique measurement identification.

Agreed, then let's keep this field and document what is its scope.

data-formats/df-001-httpt.md
bassosimone: (I picked up a slightly more compact approach but in line with what you suggested)

bassosimone:
@hellais I've addressed all comments, please take another look!

data-formats/df-000-base.md
```diff
@@ -10,135 +10,245 @@ The output of test is the composition of:
...
-    "test_helpers": {"test_specific_test_helper_name": "The address of the test helper used. (Note that the name used for the helper is not necessarily the helper name according to the bouncer, and is typically test specific; e.g. WebConnectivity currently uses 'backend' to refer to the 'web-connectivity' helper."},
+  - `0.2.1`: minor revision of the JSON format. OONI Probe CLI v3.x and
+    OONI Probe Mobile using `github.com/ooni/probe-engine`.
 
-    "test_name": "The name of the test that generated "
-        "this measurement (ex. http_requests)",
+- `id` (`string`): UUID4 identifying this specific measurement. This `id` MAY
+be used by clients to cross reference related measurements as long as they
+are submitted as part of the same report. Apart from this use case, the OONI
+backends SHOULD NOT trust this identifier to uniquely identify measurements
```
hellais: I would say rather than the OONI backends, any consumer of the data wether it's OONI or some third party.

I suggest also dropping the line on the `report_id` as it's a bit misleading. `report_id` doesn't actually uniquely identify a measurement, but rather a group of measurements so it fits a different purpose.

I also think it's important to clarify in this text that this ID is client generated.

data-formats/df-000-base.md
```diff
@@ -10,135 +10,245 @@ The output of test is the composition of:
...
+
+- `test_name` (`string`): name of the experiment in snake case. For example,
+Web Connectivity SHOULD be indicated as `"web_connectivity"`.
+
+- `test_runtime` (`float`): runtime of this specific measurement in seconds
+with arbitrary sub-seconds precision. All modern implementations, i.e.
+Measurement Kit and `github.com/ooni/probe-engine`, measure this value as
+the time elapsed since when we start measuring a specific input (or when
+we start an experiment without input) until when the measurement is
+available. This specifically means that this field does not include the
```
hellais: > until when the measurement is available

What does this mean? Available where?

bassosimone:
@hellais I've [addressed your additional requests](https://github.com/ooni/spec/compare/acd75f02520b250fe91d9c0bd797a95d936b4894..2d58032db7562f3f2abe1bbfdb0ed6441650f546), PTAL?

PR #166, opened 2019-12-02
Fix data format and use v0.3.0

### LINKED ISSUE #163: Bump data format to v0.3.0
Discussed with @FedericoCeratto while chatting about https://github.com/ooni/pipeline/issues/239

bassosimone:
As pointed out by @FedericoCeratto, using v0.2.1 for the data format
where resolver_ip is toplevel is misleading.

We do not really use the data format when processing data. Also, setting
resolver_ip inside test_keys was a Measurement Kit extension.

Notwithstanding the two facts above, we have basically removed a field
that was expected to be inside test_keys at top level.

So, using the point of view of the pipeline, this is a breaking change
even if it's possible to argue this was not because we did in fact just
remove an extension introduced by MK.

Thus, it seems more clean to use v0.3.0 from now on. This diff updates
the relevant specs to reference the v0.3.0 version. It also ensures that
we mention what happened to v0.2.1, for future reference.

Closes #163


data-formats/df-000-base.md
```diff
@@ -134,7 +138,7 @@ e.g. Measurement Kit, typically emit an empty list here. All clients using
 
 - `probe_city` (`string`; optional; deprecated): name of the city where the
```
anadahz: @bassosimone Is this still valid?: `optional and deprecated`

bassosimone: Thanks for reading this diff! Yes, it's still valid. It's `optional` because new clients do not use it and `deprecated` because we eventually want to stop advertising this field here.

anadahz: From: https://github.com/ooni/spec/pull/166/files#diff-e488a6df9676b20d84342964000b116aR58 
`Clients using v0.3.0 of this specification SHOULD NOT emit this field.`

This is a bit confusing, if this field is optional and (hopefully) deprecated why v0.3.0 clients should use it?

bassosimone: It says clients SHOULD NOT use it, and I find it consistent with the spec

anadahz: > It says clients SHOULD NOT use it, and I find it consistent with the spec

You are right, I misinterpret this statement.
My apologies for the noise.

bassosimone: No worries!

Issue #168, opened 2019-12-03
df-001-httpt.md: revamp documentation


bassosimone:
Ideally improve its quality to be on par with the base data format.

bassosimone:
This was done in Sprint 4 by me as part of my effort to do https://github.com/ooni/probe-engine/issues/90.

PR #169, opened 2019-12-03
df-001-httpt.md: document that header values may be binary

### LINKED ISSUE #164: http: document that headers may be binary
This is not explicitly stated in https://github.com/ooni/spec/blob/master/data-formats/df-001-httpt.md

bassosimone:
Closes #164


bassosimone:
I didn't bother with revamping the docs in this cycle, since there's not much time left. I have created an issue for improving this file more in the next cycle: #168

PR #175, opened 2020-01-12
ts-023-tor.md: document the new tor test

bassosimone:
* master issue: https://github.com/ooni/probe-engine/issues/226

* implementation: https://github.com/ooni/probe-engine/issues/227

### DIFF
--- a/nettests/ts-023-tor.md
+++ b/nettests/ts-023-tor.md
@@ -168,282 +168,72 @@ to more precisely classify possible `obfs4` failures.
     "_probe_engine_sanitize_test_keys": "true"
   },
   "data_format_version": "0.3.4",
-  "measurement_start_time": "2020-01-12 10:32:46",
-  "test_runtime": 3.267954277,
+  "measurement_start_time": "2020-01-16 14:41:31",
+  "test_runtime": 0.72184469,
   "probe_asn": "AS30722",
   "probe_cc": "IT",
   "probe_ip": "127.0.0.1",
-  "report_id": "20200112T103246Z_AS30722_YppBch5h3LxXgjFZ9ziVcQX89udUIcRFVbH5ar375mipxVlJVz",
+  "report_id": "20200116T144131Z_AS30722_VCY4K6dIcplLIGEQo4NnDCPyrVBMDjG99fIaNWJMuyz5gEihD5",
   "resolver_asn": "AS15169",
-  "resolver_ip": "172.253.197.2",
+  "resolver_ip": "172.217.34.4",
   "resolver_network_name": "Google LLC",
   "software_name": "miniooni",
   "software_version": "0.1.0-dev",
   "test_keys": {
-    "targets": [
-      {
+    "targets": {
+      "f372c264e9a470335d9ac79fe780847bda052aa3b6a9ee5ff497cb6501634f9f": {
         "agent": "redirect",
         "failure": null,
         "network_events": [
           {
-            "address": "109.105.109.165:10527",
-            "conn_id": 2,
-            "dial_id": 2,
+            "address": "38.229.1.78:80",
+            "conn_id": 3,
+            "dial_id": 3,
             "failure": null,
             "operation": "connect",
             "proto": "tcp",
-            "t": 1.492101
+            "t": 0.413078
           },
           {
-            "conn_id": 2,
+            "conn_id": 3,
             "failure": null,
-            "num_bytes": 3465,
+            "num_bytes": 6247,
             "operation": "write",
             "proto": "tcp",
-            "t": 1.498227
+            "t": 0.415011
           },
           {
-            "conn_id": 2,
+            "conn_id": 3,
             "failure": null,
-            "num_bytes": 2880,
+            "num_bytes": 2608,
             "operation": "read",
             "proto": "tcp",
-            "t": 1.559525
-          },
-          {
-            "conn_id": 2,
-            "failure": null,
-            "num_bytes": 1348,
-            "operation": "read",
-            "proto": "tcp",
-            "t": 1.559858
+            "t": 0.721159
           }
         ],
         "queries": null,
         "requests": null,
-        "target_address": "109.105.109.165:10527",
+        "target_address": "38.229.1.78:80",
         "target_protocol": "obfs4",
-        "tcp_connect": [
-          {
-            "conn_id": 2,
-            "dial_id": 2,
-            "ip": "109.105.109.165",
-            "port": 10527,
-            "status": {
-              "failure": null,
-              "success": true
-            },
-            "t": 1.492101
-          }
-        ],
-        "tls_handshakes": null
-      },
-      {
-        "agent": "redirect",
-        "failure": null,
-        "network_events": [
-          {
-            "address": "66.111.2.131:9001",
-            "conn_id": 1,
-            "dial_id": 1,
-            "failure": null,
-            "operation": "connect",
-            "proto": "tcp",
-            "t": 1.433779
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 243,
-            "operation": "write",
-            "proto": "tcp",
-            "t": 1.434547
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 517,
-            "operation": "read",
-            "proto": "tcp",
-            "t": 1.567884
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 489,
-            "operation": "read",
-            "proto": "tcp",
-            "t": 1.5682779999999998
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 126,
-            "operation": "write",
-            "proto": "tcp",
-            "t": 1.570798
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 51,
-            "operation": "read",
-            "proto": "tcp",
-            "t": 1.691703
-          },
-          {
-            "conn_id": 1,
-            "failure": null,
-            "num_bytes": 31,
-            "operation": "write",
-            "proto": "tcp",
-            "t": 1.692134
-          }
-        ],
-        "queries": null,
-        "requests": null,
-        "target_address": "66.111.2.131:9001",
-        "target_protocol": "or_port",
-        "tcp_connect": [
-          {
-            "conn_id": 1,
-            "dial_id": 1,
-            "ip": "66.111.2.131",
-            "port": 9001,
-            "status": {
-              "failure": null,
-              "success": true
-            },
-            "t": 1.433779
-          }
-        ],
-        "tls_handshakes": [
-          {
-            "cipher_suite": "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
-            "conn_id": 1,
-            "failure": null,
-            "negotiated_protocol": "",
-            "peer_certificates": [
-              {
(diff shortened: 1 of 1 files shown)


nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+
+The test loops through the list of targets measuring each. The
+measurement action depends on the target type:
+
+- for `dir_port` targets, the nettest will `GET` the
+`/tor/status-vote/current/consensus.z` resource using
+the HTTP protocol;
+
+- for `or_port` targets, the nettest will connect to
+the address and perform a TLS handshake;
```
hellais: Maybe we want to add that an additional thing we could do the `or_port` is also do a full Tor handshake using a tor client? This can be marked as optional.

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+
+- for `obfs4` targets, the nettest will connect to
+the address and perform an OBFS4 handshake;
+
+- otherwise, the nettest will TCP connect to the address.
```
hellais: Do we want to define some schema for the input format? I would at least specify which are the fields which could exist, even if we are not defining the exact encoding.

For example:

Each item in the input will have the following attributes:

- IP address
- Port number
- Address type, which can be one of `dir_port`, `or_port`, `tor_pt`, `tcp`
- Identifier, which is an ID used to identify this particular target in the `input` field of the measurement. In most cases it will be the fingerprint of the bridge or the hash of the fingerprint, for private bridges

In the case of `tor_pt`, we will also provide as input to the test the full bridge line, which can be something like:
```
obfs4 198.245.60.50:443 752CF7825B3B9EA6A98C83AC41F7099D67007EA5 cert=xpmQtKUqQ/6v5X7ijgYE/f03+l2/EuQ1dexjyUhh16wQlu/cpXUGalmhDIlhuiQPNEKmKw iat-mode=0 
```

and use the pluggable transport name (`obfs4` in the above example) as the value for `target_protocol` below.

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+
+```JSON
+{
+    "test_keys": {
+        "targets": []
+    }
+}
+```
+
+- `targets` (`[]Targets`): list of measurements by target
```
hellais: From a data analysis perspective it's more convenient to have these be keyed values so that it's easier to lookup the values for a given key without having to rebuild this mapping in memory.

Perhaps you can use what I call the "identifier" above as the key?

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+- `agent` (`string`): see `df-001-httpt`;
+
+- `failure` (`string`; nullable): `null` on success, string on
+error as documented in `df-007-errors.md`;
+
+- `network_events` (`[]NetworkEvent`; nullable): see `df-008-netevents.md`;
+
+- `queries` (`[]Query`; nullable): see `df-002-dnst.md`;
+
+- `requests` (`[]Transaction`; nullable): see `df-001-httpd.md`;
```
hellais: I think there is a typo, it should say `df-001-httpt.md`

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+- `target_address` (`string`): address of the target, generally expressed
+in the `1.1.1.1:555` or the `[::1]:555` or the `domain:555` forms;
+
+- `target_protocol` (`string`): protocol to access the target, which
+determines the action performed as described above;
+
+- `tcp_connect` (`[]TCPConnect`; nullable): see `df-005-tcpconnect.md`;
+
+- `tls_handshakes` (`[]Handshake`; nullable): see `df-006-tlshandshake.md`.
+
```
hellais: Maybe you also want to the output information related to the bridge metadata?

nettests/ts-023-tor.md
bassosimone: lol, you're right

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+better understand the reason for failure.
+
+## Example output sample
+
+```JSON
+{
+  "annotations": {
+    "_probe_engine_sanitize_test_keys": "true"
+  },
+  "data_format_version": "0.3.4",
```
hellais: I think you are missing the `input` key, which should be set to `null`

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+
+- `queries` (`[]Query`; nullable): see `df-002-dnst.md`;
+
+- `requests` (`[]Transaction`; nullable): see `df-001-httpd.md`;
+
+- `target_address` (`string`): address of the target, generally expressed
+in the `1.1.1.1:555` or the `[::1]:555` or the `domain:555` forms;
+
+- `target_protocol` (`string`): protocol to access the target, which
+determines the action performed as described above;
```
hellais: In the case of obfs4proxy, I think we should also capture the log output of obfs4proxy as this might include useful information related to at which point during the handshake it failed. Ideally we would make some changes to obfs4proxy in order to even capture some amount of state to identify during which part of the handshake it failed (ex. during the 3rd roundtrip).

Perhaps @nullhypothesis has some thoughts on this too.

nettests/ts-023-tor.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+# Privacy considerations
+
+This nettest does not provide anonymity. An adversary can observe that
+the user is connecting to Tor servers and using obfs4.
+
+# Packet capture considerations
+
+This test does not capture packets by default.
```
hellais: Maybe this section should be renamed something different from packet capture consideration, but something more generic like "data minimisation considerations".

For this test in particular one thing we should document and be aware of is the fact that some of the bridges we could end up testing may not be public. In this case we need to be careful to:

1. Sanitise the identifier inside of the target list replacing it with a double hash
2. Replace any reference to the bridge IP address and port inside of the `tcp_connect`, `requests` and `queries` keys with a sanitised value (ex. `127.2.3.4`)
3. Search and replace the bridge IP address, shared secret and fingerprint inside of the logs (if we do logging of obfs4proxy) to prevent leakage of it

nettests/ts-023-tor.md
bassosimone: I think we should write that down as a possible improvement. I am reluctant to specify this possibility until I have a proof of concept that allows me to sketch out the expected output. I agree it should be optional. I also think if we should look into using and/or adapting bridgestrap to do this kind of work.

nettests/ts-023-tor.md
bassosimone: This is the easiest way to represent this information:

```JSON
// TorTarget is a target for the tor experiment.
type TorTarget struct {
	// Address is the address of the target.
	Address string `json:"address"`

	// Params contains optional params for, e.g., pluggable transports.
	Params map[string][]string `json:"params"`

	// Protocol is the protocol to use with the target.
	Protocol string `json:"protocol"`
}
```

It is in fact what I am using inside the implementation. I'd rather not have `tor_pt` as type but have directly the kind of pluggable transport that I should invoke. How about we always hash the fingerprint and we never publish the target so we're more robust? Is that annoying?

Regarding the bridge line, I think it may be useful specifying that we can pass one to the command line and we can parse it like the data structure above. Given that the type is the first token, we can uniformly treat all the other kind of entries in the same way.

nettests/ts-023-tor.md
bassosimone: Sure

nettests/ts-023-tor.md
bassosimone: I am struggling to find logging statements in the obfs4 code we're using 🤷‍♀

bassosimone: Regarding making changes to identify when it fails, yes, maybe as a future enhancement?

nettests/ts-023-tor.md
bassosimone: So, the (hash of the) fingerprint is going to be the primary key in a map. If the key is not sensitive, sure!

nettests/ts-023-tor.md
bassosimone: I have made it optional recently. Any sensible consumer shall check for that case anyway.

nettests/ts-023-tor.md
bassosimone: The "packet capture consideration" section is a somewhat standard one. The content I wrote is not in line with the content it should actually contain. I'll change the title, then!

nettests/ts-023-tor.md
hellais: Sounds good.

nettests/ts-023-tor.md
hellais: > How about we always hash the fingerprint and we never publish the target so we're more robust? Is that annoying?

I think it's a little bit annoying, but I agree it's going to be more robust, so I say we go for it.

We should then be sure to specify in the data format returned to the client that the fingerprint is in a separate field.

For example in your above data model you do not have a specific field for the fingerprint, we should have it.

nettests/ts-023-tor.md
hellais: I guess in light of your comment above, we should probably emit only the hashed fingerprint and omit the key.

nettests/ts-023-tor.md
hellais: @FedericoCeratto are you aware of this? I am worried that the legacy pipeline may not handle cases in which the `input` field is missing. Have we tested this?

nettests/ts-023-tor.md
hellais: Yes perhaps we can make it a future enhancement.

nettests/ts-023-tor.md
hellais: > The "packet capture consideration" section is a somewhat standard one

yes I know. I was suggesting to redefine the standard section to be this more generic thing which applies in fact to most tests and also includes packet capture considerations.

nettests/ts-023-tor.md
bassosimone: > For example in your above data model you do not have a specific field for the fingerprint, we should have it.

Right!

nettests/ts-023-tor.md
bassosimone: Maybe this is a good moment to test it. The code where the `input` is not optional has not been released in production yet. I was relatively confident because it seems to me that the centrifugation code considers the `input` as nullable, though.

nettests/ts-023-tor.md
bassosimone: This has been fixed, I will push a fix soon.

nettests/ts-023-tor.md
bassosimone: No further action required here, once we're using the fingerprint as the key.

nettests/ts-023-tor.md
bassosimone: This has been done, I will push the updated spec soon.

nettests/ts-023-tor.md
bassosimone: This has been done and I will push it soon.

nettests/ts-023-tor.md
bassosimone: I have created follow-up issues for this action: https://github.com/ooni/backend/issues/280, and https://github.com/ooni/probe/issues/971

nettests/ts-023-tor.md
bassosimone: I have listed this action as a possible future improvement inside the spec.

nettests/ts-023-tor.md
bassosimone: In the next cycle we should follow-up with a discussion regarding how to sanitize the data (and we also need to spec out how to create identifiers for inputs without fingerprint). Since I am now realizing there is a bunch of "todo-next" issues, I am leaning towards listing all the limitations in the current iteration of the spec, and creating follow-up issues for each of them such that we can follow-up in the next cycle. (Meanwhile, since the bulk of this new experiment is now here, I am also leaning towards merging it and publishing a new release candidate for `probe-cli`, so we can move forward the work that depends on this.)

bassosimone:
I have created a bunch of issues (ooni/probe-engine#270, ooni/probe-engine#271, ooni/probe-engine#272, and ooni/probe-engine#273) to address known limitations of this specification in the moment in which we're merging it. I am now going to merge.

PR #177, opened 2020-02-18
Explain how we derive the id of an input.

NullHypothesis:
See the following PR for more context:
https://github.com/ooni/sysadmin/pull/424


hellais:
🐳

PR #179, opened 2020-03-04
Fix the typos.

tomac4t:
Pointed out by dcf.

### DIFF
--- a/nettests/ts-007-http-invalid-request-line.md
+++ b/nettests/ts-007-http-invalid-request-line.md
@@ -56,7 +56,7 @@ Performs a request with a HTTP method that is a random 4
 character string. e.g.
 
 ```
-request_line = randomSTR(4) + " / HTTP/1.1\n\r"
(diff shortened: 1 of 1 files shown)


bassosimone:
We previously discussed this issue and it was not clear whether we were okay with applying that change or not, as you can read here: https://github.com/ooni/probe-engine/issues/246.

Cc: @hellais

hellais:
I think the best thing is probably to measure both. As you said we have discussed it in the past and it's probably useful to not change the thing we are measuring, but rather do it in addition to it.

We also know that even with the inverted CRLF, we are still triggering middlebox detection and we don't know if that is going to be the case if we make the order correct.

In any case the spec is not wrong, because it's in fact saying what we do in the test :P

tomac4t:
> In any case the spec is not wrong, because it's in fact saying what we do in the test :P

I think it could close this PR without the problem. :smile:

PR #202, opened 2020-10-14
RiseupVPN spec

KwadroNaut:
This spec goes along with https://github.com/ooni/probe-engine/pull/978
Please comment if, for whatever reason, this should be changed.

### DIFF
--- a/nettests/ts-026-riseupvpn.md
+++ b/nettests/ts-026-riseupvpn.md
@@ -1,0 +1,134 @@
+# Specification version number
+
+2020-09-14-000
+
+# Specification name
+
+RiseupVPN
+
+# Test preconditions
+
+* An internet connection with functional DNS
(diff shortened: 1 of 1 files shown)


bassosimone (approved):
Thank you so much for creating an experiment for LEAP VPNs. I am going to accept this PR with the following changes authored by myself on top of the original diff:

1. generic editorial changes to make the text a bit more consistent with other specs

2. specific changes to ensure names inside JSON documents are lowercase (mainly because this is what we do in most other existing specs), so I am trying to keep consistency here

I have been thinking about whether the riseupvpn name is correct in this context, especially considering the explicitly stated aspiration to extend this experiment to other LEAP providers.

I would therefore like to encourage you to think at whether a more general name is possible.

That said, I also feel like it's okay to go ahead with the experiment with the current name and continue the conversation once we have merged some experimental, running code as well.

PR #210, opened 2021-02-15
Add signal test specification

hellais:
xref: https://github.com/ooni/probe-cli/pull/230

Reference issue: https://github.com/ooni/probe-engine/issues/278


bassosimone (approved):
:whale:

PR #211, opened 2021-02-16
reduce false positive rate in RiseupVPN tests

cyBerta:
introduce transport_status for a better filtering of false positive RiseupVPN nettest failures
part of https://github.com/ooni/probe/issues/1354

### DIFF
--- a/nettests/ts-026-riseupvpn.md
+++ b/nettests/ts-026-riseupvpn.md
@@ -71,7 +71,7 @@ If all parts of the API are functional and reachable then we write:
 ## RiseupVPN gateways test
 
 If the provider API is reachable, it provides a JSON-file which contains the IP addresses and capabilites of the VPN gateways. The reachability of gateways will be tested depending on their capabilities as described by the provider (ports, OpenVPN, obfs4) by performing TCP handshakes. If a TCP handshake fails we assume the corresponding port and transport of that gateway to be blocked and add it to the list of failing gateways. 
-We consider a transport to be blocked if it was impossible to connect to any gateway port providing that transport.
+We consider a transport to be accessible if it was possible to connect at least to one gateway port providing that transport.
 
 Example output for reported blocked gateways:
 
--- a/nettests/ts-026-riseupvpn.md
+++ b/nettests/ts-026-riseupvpn.md
@@ -1,6 +1,6 @@
 # Specification version number
 
-2020-10-13-000
+2021-02-15-000
 
 # Specification name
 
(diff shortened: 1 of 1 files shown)


cyBerta:
@KwadroNaut could you have a look at it, too, please?

KwadroNaut:
It's not clear for me, from the json, whether you want to flag it as blocked/broken/..when openvpn AND obfs4 are blocked or OR?

cyBerta:
> It's not clear for me, from the json, whether you want to flag it as blocked/broken/..when openvpn AND obfs4 are blocked or OR?

I've improved the json. Each transport has an own status flag.

cyBerta:
@bassosimone I think it's a good time now to merge the spec as well, since the implementation already changed accordingly.

nettests/ts-026-riseupvpn.md
```diff
@@ -70,12 +70,17 @@ If all parts of the API are functional and reachable then we write:
 
 ## RiseupVPN gateways test
 
-If the provider API is reachable, it provides a JSON-file which contains the IP addresses and capabilites of the VPN gateways. The reachability of gateways will be tested depending on their capabilities as described by the provider (ports, OpenVPN, obfs4) by performing TCP handshakes. If a TCP handshake fails we can assume the corresponding port and transport of that gateway to be blocked.
+If the provider API is reachable, it provides a JSON-file which contains the IP addresses and capabilites of the VPN gateways. The reachability of gateways will be tested depending on their capabilities as described by the provider (ports, OpenVPN, obfs4) by performing TCP handshakes. If a TCP handshake fails we assume the corresponding port and transport of that gateway to be blocked and add it to the list of failing gateways. 
+We consider a transport to be blocked if it was impossible to connect to any gateway port providing that transport.
```
bassosimone, line 74: Given this statement:

> We consider a transport to be blocked if it was impossible to connect to any gateway port providing that transport.

it seems to me that the following code:

```Go
func (tk *TestKeys) updateTransportStatus(openvpnGatewayCount int, obfs4GatewayCount int) {
	failingOpenvpnGateways, failingObfs4Gateways := 0, 0
	for _, gw := range tk.FailingGateways {
		if gw.TransportType == "openvpn" {
			failingOpenvpnGateways++
		} else if gw.TransportType == "obfs4" {
			failingObfs4Gateways++
		}
	}
	if failingOpenvpnGateways < openvpnGatewayCount {
		tk.TransportStatus["openvpn"] = "ok"
	} else {
		tk.TransportStatus["openvpn"] = "blocked"
	}
	if failingObfs4Gateways < obfs4GatewayCount {
		tk.TransportStatus["obfs4"] = "ok"
	} else {
		tk.TransportStatus["obfs4"] = "blocked"
	}
}
```

does not implement the statement. In the code above you set blocked if the number of failing gateways is `>=` than the total number of gateways. My understanding of the statement is that you should set the transport blocked if _at least one_ gateway (i.e. "any") is blocked for the transport.

Unless I'm misinterpreting here, your aim is to make false positive less likely. So, I suppose the code is correct and the specification should instead say "to all gateway ports"?

bassosimone (commented):
I have just a question about a sentence that perhaps I don't fully understand and perhaps instead does not match the implementation. Otherwise, the spec change looks good!

Thank you for your great work!

nettests/ts-026-riseupvpn.md
cyBerta: you're right, the wording is wrong. I confused any with none :upside_down_face:. I'll fix that

bassosimone (approved):
Thank you!

PR #218, opened 2021-06-18
doc: document ts-030-torsf

bassosimone:
See https://github.com/ooni/probe/issues/1565

### DIFF
--- a/nettests/ts-030-torsf.md
+++ b/nettests/ts-030-torsf.md
@@ -1,0 +1,105 @@
+# Specification version number
+
+2021-06-18-001
+
+# Specification name
+
+`torsf` (Tor using snowflake)
+
+# Test preconditions
+
+* An internet connection
+* The `tor` binary installed on the system
+
+# Expected impact
+
+Ability to detect if `tor` would bootstrap using the Snowflake
+pluggable transport (PT) within a reasonable timeout.
+
+# Expected inputs
+
+None
+
+# Test description
+
+This test starts `tor` using the following command line arguments:
+
+- `UseBridges` `1`
+- `ClientTransportPlugin` `'snowflake socks5 127.0.0.1:PORT'`
+- `Bridge` `'snowflake ADDR_PORT FINGERPRINT'`
+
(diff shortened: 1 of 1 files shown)


nettests/ts-030-torsf.md
```diff
@@ -0,0 +1,105 @@
...
+## Parent data format
+
+* none
+
+## Semantics
+
+```JSON
+{
+    "bootstrap_time": 1.1,
+    "failure": null
```
hellais, line 59: Since the tor daemon is going to be outside of our control, do we perhaps want to include the tor version as well as part of the test-keys?

nettests/ts-030-torsf.md
```diff
@@ -0,0 +1,105 @@
...
+}
```
hellais, line 60: Can we also include the percentage of tor bootstrap when it fails? I think this is helpful to determine at which point the potential blocking occurs.

Moreover, is there some additional information we can extract from snowflake itself which gives us more information about when and how it's failing? I am thinking maybe we can record the raw logs of snowflake and dump them inside of the test_keys.

nettests/ts-030-torsf.md
bassosimone: This is a good suggestion, thank you! I am going to create a new issue detailing the next steps as far as this specific test is concerned, and I am definitely gong to list this one as part of that!

nettests/ts-030-torsf.md
bassosimone: Yes, indeed! I'd like to collect all these ideas in a follow-up issue. We have not still figured out completely how to integrate tor and we need to decide whether to use https://github.com/ipsn/go-libtor. I would like to first look into that and figure out what is easily doable, then circle back and update the spec and its implementation.

bassosimone: Created epic for future work: https://github.com/ooni/probe/issues/1686

nettests/ts-030-torsf.md
bassosimone: Created epic for future work: https://github.com/ooni/probe/issues/1686

PR #219, opened 2021-08-09
th-007-websteps: websteps test helper spec

bassosimone:
See https://github.com/ooni/probe/issues/1729 and https://github.com/ooni/probe-cli/pull/431

### DIFF
--- a/backends/th-007-nwcth.md
+++ b/backends/th-007-nwcth.md
@@ -14,7 +14,7 @@ deriving from the input URL that OONI Probe should test.
 
 This discovery includes finding out HTTPS URLs corresponding to HTTP URLs
 in case of 301 redirections. It also includes discovering whether there are
-QUIC endpoints associated to HTTPS URLs.
+HTTP/3 endpoints associated to HTTPS URLs.
 
 Because we perform this discovery, we also cover the case where a Probe
 cannot continue to perform a measurement, because it receives a `NXDOMAIN`
@@ -308,7 +308,7 @@ the `CtrlResponse` will be as follows in this case:
   "urls": [{
     "url": "http://nonexistent.com",
     "dns": {
-      "failure": "dns_nxdomain_error"
+      "failure": "dns_nxdomain_error",
       "addrs": []
     },
     "endpoints": []
@@ -322,7 +322,7 @@ helper will then discover all the URLs that matter.
 The reason why we check for a nonexisting domain immediately
 is that around 1% of the domains in the test lists are not
 registered anymore (according to [ooni/probe#1727](
-https://github.com/ooni/probe/issues/1727). Given the burden
+https://github.com/ooni/probe/issues/1727)). Given the burden
 of keeping the test lists _always_ up to date, the test
 helper provides for this relatively common error case and
 includes code to handle it in the most graceful way.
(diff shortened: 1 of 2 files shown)


backends/th-007-nwcth.md
```diff
@@ -0,0 +1,227 @@
...
+1. adding logic to the test helper to choose whether it is smart
+to continue processing due to other considerations. For example, if
+we're redirected from `http://facebook.com/` to `https://facebook.com/`,
+it may be reasonable to stop the measurement instead of being
+further redirected to, say, `https://facebook.com/en_US/login.php`.
+
+2. whether the amount of information included in each `EndpointMeasurement{}`
+is such that the OONI Probe can unambiguously redo the same operation.
+
+3. because [recent measurements](https://github.com/ooni/probe/issues/1727) show that we can perform most
```
kelmenhorst, line 220: the test helper forwards cookies to redirect requests via the http.CookieJar though?

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,227 @@
...
+* _Author_: sbs
+* _Version_: 2021-08-09-001
+* _Status_: alpha
+
+This document describes a draft specification for the new web connectivity test
+helper. We tentatively expose this new API as `/api/unstable/wcth`.
+
+For brevity, we will not delve into rationale and other menial details here.
+
+## Request message
```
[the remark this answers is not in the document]
bassosimone: Good point! I just did that, thanks!

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,227 @@
...
+
+The request message is exactly as in the current test helper:
+
+```
+{
+  "http_request": URL"",
+  "http_request_headers": map[string][]string{},
+  "tcp_connect": []Endpoint""
+}
```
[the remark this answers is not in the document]
bassosimone: Yes, we should. Actually, we should perhaps wrap the data structure we use in the implementation to be like:

```
type CtrlRequestWrapper struct {
  HTTPCookieJar http.CookieJar
  CtrlRequest CtrlRequest
}
```

and then we should explicitly read just the `CtrlRequest` part in `json.Marshal` and init `HTTPCookieJar` separately. This seems the most robust way to separate remote provided data and local data.

(The `json:"-"` is also a valid approach but it's probably more tidy to achieve a better separation by using two different data structures rather than just using annotation magic.)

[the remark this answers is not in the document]
kelmenhorst: yes this seems better. I will do so now.

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+that covers the client side implementation of the new test.
+
+## Semantics
+
+This section defines the messages of the New Web
+Connectivity Test Helper protocol.
+
+### Endpoint
+
+Before discussing messages, we need to define _endpoint_, which is
```
kelmenhorst, line 37: .. a TCP or UDP endpoint.
In terms of syntax, this means ...
(maybe it would be more precise to put it like this?)

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+
+where:
+
+- `url` is mandatory and contains the string-serialized URL we should measure;
+
+- `headers` is an optional map from string to a list of strings containing the headers
+the test helper should include when measuring (which headers from this map will be used
+in specified more in detail below);
+
+- `addrs` is an optional list of string-serialized IP addresses for the
```
kelmenhorst, line 79: IIRC it's a list of endpoints, not just addresses (?)

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+
+The order of the `CtrlResponse.urls` field is the order in which the
+test helper wants the probe to perform measurements.
+
+#### DNSMeasurement
+
+The `DNSMeasurement` message contains these fields:
+
+```
+DNSMeasurement = {
```
kelmenhorst, line 147: Should `DNSMeasurement` also include the domain name that is to be resolved? In case of an error, you can directly see which domain name could not be resolved then.

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+`df-007-errors.md` document for more information);
+
+- `addrs` is a possibly-empty list of IP addresses returned by the DNS lookup.
+
+#### EndpointMeasurement
+
+The `EndpointMeasurement` messsages is either `HTTPMeasurement` or `H3Measurement`:
+
+```
+EndpointMeasurement = HTTPEndpointMeasurement | H3EndpointMeasurement
```
kelmenhorst, line 165: I propose to distinguish between `HTTPEndpointMeasurement` and `HTTPSEndpointMeasurement`. Otherwise, `HTTPEndpointMeasurement` contains an empty `TLSHandshake` field, so the value of `Failure` is false, which can be confusing because it looks exactly like a successful TLSHandshake.

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+
+#### QUICHandshakeMeasurement
+
+`QUICHandshakeMeasurement` is (currently) an alias for `TLSHandshakeMeasurement`.
+
+```
+QUICHandshakeMeasurement = TLSHandshakeMeasurement
+```
+
+#### HTTPRequestMeasurement
```
kelmenhorst, line 259: I propose to replace `HTTPRequestMeasurement` with `HTTPRoundtripMeasurement`, where `HTTPRoundtripMeasurement` has the following fields:
```
HTTPRequestMeasurement = {
  "request": HTTPRequest,
  "response": HTTPResponse,
}
```
```
HTTPRequest = {
"headers": {},
}
```
(rationale: The probe can use `HTTPRequest` to set the correct headers in its own measurement.)
```
HTTPResponse = {
  "body_length": 0,
  "failure": null | "",
  "headers": {},
  "status_code": 0,
}
```

backends/th-007-nwcth.md
```diff
@@ -0,0 +1,623 @@
...
+includes discovering HTTP/3 endpoints. We discover
+them using `Alt-Svc`. In the future, we should
+also consider using DNS SVCB.
+
+This HTTP client will use the `accept`, `accept-language`,
+and `user-agent` headers provided by the client in the
+`headers` field of the `CtrlRequest`. All other headers
+will be silently ignored.
+
+The output of this step consists of three lists of requests, for
```
kelmenhorst, line 369: What is the advantage of having three separate lists? Instead one list that contains, for example:
`[http://google.com, https://google.com, h3://google.com, https://www.google.com, h3://www.google.com]`

backends/th-007-nwcth.md
bassosimone: Sure!

backends/th-007-nwcth.md
bassosimone: Yes, I changed it to a list of IP addresses because it's closer to what the probe actually provided and because it allows for more uniform treatment of the Probe and the test helper DNS results.

(Ouch, I had a note mentioning this was changed but it seems I actually removed it from the text.)

backends/th-007-nwcth.md
bassosimone: Good idea!

backends/th-007-nwcth.md
bassosimone: Good idea!

backends/th-007-nwcth.md
bassosimone: Good idea!

backends/th-007-nwcth.md
bassosimone: There is no advantage except that it's easier to explain. In a bunch of places in this spec I mention how things should be in a lax sense: any implementation that is isomorphic to that is fine. Should I make this explicit?

bassosimone: Something like this

```This section describes algorithms and data structures to implement the test helper. An implementation is free to choose any equivalent set of algorithms and data structures as long as the end result is functionally equivalent to what we describe in here.```

when entering into the algorithms section should do.

backends/th-007-nwcth.md
bassosimone: ✔️

backends/th-007-nwcth.md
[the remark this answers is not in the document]
bassosimone: 👍

backends/th-007-nwcth.md
bassosimone: ✔️

backends/th-007-nwcth.md
bassosimone: ✔️

backends/th-007-nwcth.md
bassosimone: ✔️

backends/th-007-nwcth.md
bassosimone: ✔️

bassosimone (commented):
Thank you for your review @kelmenhorst! I have applied the suggested changes and the document is now ready for another round of review!

kelmenhorst (approved):
We discussed this spec and agreed upon the proposed data format and the principles of the algorithm. The minor differences between the implementation and the spec are not relevant, as the applied algorithm is the same.

bassosimone:
Let's merge this version of the spec into the main branch. We will continue tweaking and changing the spec for quite some time, but I think it is time to call this part of the work done and move on.

PR #230, opened 2022-02-03
ts-031-quicping: quicping spec

kelmenhorst:
This refers to https://github.com/ooni/probe/issues/1994 and is related to https://github.com/ooni/probe-cli/pull/677.

The spec describes the new `quicping` experimental network test.

### DIFF
--- a/nettests/ts-031-quicping.md
+++ b/nettests/ts-031-quicping.md
@@ -29,7 +29,8 @@ The experiment takes in as input a single domain name or IP address, or a list o
 
 The following options can be configured (for now by using the `miniooni` client):
 - target port, option `Port` (`int64`, default: 443)
-- number of ping repetitions, option `Repetitions` (`int64`, default: 10)
+- number of ping requests that `quicping` will send, option `Repetitions` (`int64`, default: 10)
+    - `quicping` does *not* stop when it successfully receives a response but always sends `Repetitions` times of ping requests
 - number of milliseconds before read timeout, option `Timeout` (`int64`, default: 5000).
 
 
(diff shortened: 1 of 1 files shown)


nettests/ts-031-quicping.md
```diff
@@ -0,0 +1,173 @@
...
+
+The following options can be configured (for now by using the `miniooni` client):
+- target port, option `Port` (`int64`, default: 443)
+- number of ping repetitions, option `Repetitions` (`int64`, default: 10)
+- number of milliseconds before read timeout, option `Timeout` (`int64`, default: 5000).
+
+
+
+# Test description
+
```
hellais, line 38: It's unclear to me how the value of `repetitions` impacts the test logic. Will it stop performing pings after it successfully receives a response or will it always perform that number of repetitions?

If it's the former, maybe `repetitions` could be renamed to `retries`?

In any case it should be explicitly mentioned in the description of the test.

nettests/ts-031-quicping.md
```diff
@@ -0,0 +1,173 @@
...
+}
+```
+
+where:
+
+- `domain` (`string`) is the domain of the target host
+
+- `pings` is the list of ping operations:
+    - `conn_id_dst` (`[]byte`) is the used destination connection ID
+    - `conn_id_src` (`[]byte`) is the used source connection ID
```
hellais, line 95: Maybe it should be specified here that it's base64 encoded.

nettests/ts-031-quicping.md
kelmenhorst: It's the latter, so it will always perform that number of repetitions. You're right, I will make that more explicit.

bassosimone:
LGMT, thanks! 🐳

PR #234, opened 2022-04-20
df-001-httpt.md: clarify socksproxy field meaning

linpeiyu164:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1947
- [ ] related ooni/probe-cli pull request

Location of the issue tracker: https://github.com/ooni/probe

## Description

The `socksproxy` field is only set when a nettest uses a SOCKS proxy. This is different from the SOCKS proxy used to contact the backend. Clarifications have been added to `df-001-httpt.md`.

### DIFF
--- a/data-formats/df-001-httpt.md
+++ b/data-formats/df-001-httpt.md
@@ -24,7 +24,7 @@ follow redirects, set to `redirect` otherwise.
 - `request` (`[]Transaction`): list of transaction objects. See below.
 
 - `socksproxy` (`string`; optional): address of the SOCKS proxy being
-used. Omit or set to `null` if no SOCKS proxy is being used. The format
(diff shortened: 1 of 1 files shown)


data-formats/df-001-httpt.md
```diff
@@ -6,7 +6,7 @@ code. See this directory's [README](README.md) for the basic concepts.
 
 | Name       | `httpt` |
 |------------|---------|
-| Version    | 0       |
+| Version    | 1       |
```
bassosimone, line 9:
```suggestion
| Version    | 0       |
```

I don't think there's need to bump the version number because this change is not breaking anything that previously existed but rather it's just clarifying the scope of a specific field.

bassosimone (approved):
Looks good, thank you! I'd just not change the version number because there's no need to do that, since we're just clarifying the meaning of a field.

PR #244, opened 2022-06-02
Add browser_web spec

hellais:
This is a draft spec for ooni/probe-web: https://ooni.github.io/probe-web/

### DIFF
--- a/nettests/ts-036-browser_web.md
+++ b/nettests/ts-036-browser_web.md
@@ -62,11 +62,11 @@ a TLS MITM would be visible as a failure.
 
 Moreover, if the user has enabled [strict enhanced tracking
 protection](https://support.mozilla.org/en-US/kb/enhanced-tracking-protection-firefox-desktop#w_strict-enhanced-tracking-protection)
-on Firefox, requests to certain sites (ex. twitter.com or facebook.com) will
+on Firefox, requests to [certain sites](https://github.com/disconnectme/disconnect-tracking-protection/blob/master/services.json) (ex. twitter.com or facebook.com) will
 fail and appear to be a result of blocking.
-The workaround is to detect if the user has enabled this setting and if so
-request that they add a whitelist of the site hosting the `browser_web`
-experiment.
+Firefox users should be asked, if they have strict enhanced tracking protection
+enabled, to add a whitelist rule for the website hosting the experiment,
+however this is not a guarantee that they will run the test with it on.
 
 # Expected output
 
--- a/nettests/ts-036-browser_web.md
+++ b/nettests/ts-036-browser_web.md
@@ -58,6 +58,14 @@ content is consistent with what we would expect to see. It's therefore advisable
 to only use as testing targets HTTPS websites as an inconsistent DNS response or
 a TLS MITM would be visible as a failure.
 
+Moreover, if the user has enabled [strict enhanced tracking
+protection](https://support.mozilla.org/en-US/kb/enhanced-tracking-protection-firefox-desktop#w_strict-enhanced-tracking-protection)
+on Firefox, requests to certain sites (ex. twitter.com or facebook.com) will
+fail and appear to be a result of blocking.
+The workaround is to detect if the user has enabled this setting and if so
+request that they add a whitelist of the site hosting the `browser_web`
+experiment.
+
 # Expected output
 
 ## Parent data format
--- a/nettests/ts-036-browser_web.md
+++ b/nettests/ts-036-browser_web.md
@@ -41,6 +41,8 @@ This test will produce as a final result the time it took to load the resource
 API](https://developer.mozilla.org/en-US/docs/Web/API/Performance)) and whether the
 result was successful.
 
+## Caveats
+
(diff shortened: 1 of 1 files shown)


nettests/ts-036-browser_web.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Expected inputs
+
+This experiment takes in input a list of URLs.
+
+There are some limitations in what kinds of URLs can be measured due to specific
+server-side CORS configuration.
+
+In particular, if the URL we are trying to measure is setting [CORP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)),
+it will always show up as a failure, making the measurement unreliable.
```
bassosimone, line 27: Please, document the Firefox option that was making a bunch of my measurements fail as well. There should be enough information to document the issue in our Slack logs.

nettests/ts-036-browser_web.md
```diff
@@ -0,0 +1,110 @@
...
+the test a web page is visited and after an informed consent procedure, it is
+possible to run the experiment.
+
+The test uses the [fetch
+API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) in an attempt
+to measure if certain target URLs are being blocked.
+
+This test will produce as a final result the time it took to load the resource
+(using the [performance
+API](https://developer.mozilla.org/en-US/docs/Web/API/Performance)) and if the
```
bassosimone, line 41:
```suggestion
API](https://developer.mozilla.org/en-US/docs/Web/API/Performance)) and whether the
```

bassosimone (changes requested):
Please, document the issue with Mozilla Firefox 🙏

nettests/ts-036-browser_web.md
```diff
@@ -0,0 +1,118 @@
...
+The test uses the [fetch
+API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) in an attempt
+to measure if certain target URLs are being blocked.
+
+This test will produce as a final result the time it took to load the resource
+(using the [performance
+API](https://developer.mozilla.org/en-US/docs/Web/API/Performance)) and whether the
+result was successful.
+
+Because the information collected by this test is not very detailed some extra
```
bassosimone, line 44:
```suggestion
## Caveats

Because the information collected by this test is not very detailed some extra
```

nettests/ts-036-browser_web.md
```diff
@@ -0,0 +1,118 @@
...
+  "load_time_ms": 404.40000000596046
+}
+```
+
+where:
+
+* `result` (`string`) will be `ok` in the case of a successful measurement,
+  while it will be `error` in the case of a failure;
+* `load_time_ms` (`float`) indicates the number of milliseconds it took to load
+  the requested resource;
```
bassosimone, line 89:
```suggestion
  the requested resource.
```

nettests/ts-036-browser_web.md
```diff
@@ -0,0 +1,118 @@
...
+to only use as testing targets HTTPS websites as an inconsistent DNS response or
+a TLS MITM would be visible as a failure.
+
+Moreover, if the user has enabled [strict enhanced tracking
+protection](https://support.mozilla.org/en-US/kb/enhanced-tracking-protection-firefox-desktop#w_strict-enhanced-tracking-protection)
+on Firefox, requests to certain sites (ex. twitter.com or facebook.com) will
+fail and appear to be a result of blocking.
+The workaround is to detect if the user has enabled this setting and if so
+request that they add a whitelist of the site hosting the `browser_web`
+experiment.
```
bassosimone, line 67: I am failing to understand the purpose of this sentence. Should I read it as "the implementation SHOULD detect whether the user..." or "the implementation MUST detect whether the user...". Or is it more like "we don't currently handle this problem and a future version of the experiment SHOULD handle this failure so you should be careful when analyzing the results". So, I'd invest a bit more time to clarify which case we are in (or, if it's another different case, to spell out exactly which case this is).

hellais: I did a little bit of research into this, but as far as I can tell, there isn't really a very reliable way to detect it.

The best technique I have come up with is to make a request to a large enough sample of what firefox is known to block (they use this list: https://github.com/disconnectme/disconnect-tracking-protection/blob/master/services.json) and if they all fail very quickly, then we consider it to be on.
The issue is that this may lead to false positives, because if all these sites are indeed being blocked, we would like to be able to measure it, so even in this case we probably still want to allow a manual override.

To be honest I think the best thing we can do is display a message in the beginning of the test telling firefox users that if they have enabled enhanced tracking protection they should whitelist probe-web, but we don't have any guarantee that they are going to do it, so this needs to be taken into account when analysing the data.

bassosimone (approved):
🐳

nettests/ts-036-browser_web.md
bassosimone: Yeah, this strikes me as a fundamental test limitation. Thanks for explaining!

PR #249, opened 2022-06-17
Add initial spec for OONI run

aanorbel:
## Description

Spec for OONI Run Service

### DIFF
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -570,7 +570,8 @@ is more reliable, yet it has the tradeoff of not allowing us to display a web
 page when the user does not have the app installed. As such the recommended
 strategy is to encourage users to share the custom URL prefix OONI Run link, but
 on the web page itself, in the event that the app did not handle the deep link,
-have a link to the custom prefix approach to "force" the opening of the app.
+have a link to the custom prefix approach to "force" the opening of the app
+(similar to how OONI Run works now).
 
 As such we recommend using the following addresses for OONI Run link and OONI Run descriptor URLs:
 
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -1,4 +1,4 @@
-# OONI Run specification
+# OONI Run v2 specification
 
 * author: Norbel Ambanumben, Arturo Filastò
 * version: 2023.02.09
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -102,7 +102,7 @@ An OONI Run link descriptor is a JSON file with the following semantics:
   "name_intl": {
     "it": "Il nome del test in italiano",
   },
-  
+
   // TODO: recommend a maximum length for this field
   "short_description": "(optional) `string` short_description for the OONI Run link.",
 
@@ -115,20 +115,20 @@ An OONI Run link descriptor is a JSON file with the following semantics:
   "description_intl": {
     "it": "La descrizione del test in italiano"
   },
-  
+
   "icon": "(optional) `string` the ID of any icon part of the OONI icon set",
 
   "author": "(optional) `string` name of the creator of this OONI Run link",
 
   // `array` provides a JSON array of tests to be run.
   "nettests":[{
-  
+
     // (optional) `array` provides a JSON array of tests to be run.
     "inputs": [
       "https://example.com/",
       "https://ooni.org/"
     ],
-    
+
     // (optional) `map` options arguments provided to the specified test_name
     "options": {
       "HTTP3Enabled": true
@@ -137,27 +137,35 @@ An OONI Run link descriptor is a JSON file with the following semantics:
     // (optional) `map` settings which are sent to probe_services for retrieving the inputs vector
     // It's possible to reference user configured settings by using accessing the 
     // $settings special variable.
-    // In particular the content of the test_settings will be sent to the /api/v1/check-in call nested 
+    // In particular the content of the backend_options will be sent to the /api/v1/check-in call nested 
     // under the relative test_name.
-    "test_settings": {
+    "backend_options": {
       "category_codes": "$settings.category_codes"
     },
 
     // (optional) `bool` indicates if this test should be run as part of autoruns. Defaults to true.
     "is_background_run_enabled": true,
- 
+
     // (optional) `bool` indicates if this test should be run as part of manual runs. Defaults to true.
     "is_manual_run_enabled": true,
 
     "test_name": "web_connectivity"
   }, {
     "test_name": "openvpn",
-    "test_settings": {
+    "backend_options": {
        "provider": "riseupvpn"
     },
   }]
 }
 ```
+
+In particular values starting with the `$settings` prefix should be mapped
+based on user-configured preferences in the app.
+
+Currently the following `$settings` are defined:
+
+* `$settings.category_codes` should be replaced with the list of category codes
+  the user has enabled (ex. `["HUMR", "NEWS"]`.
 
 Based on the above specification it would be possible to re-implement the cards for the OONI Probe
 dashboard as follows.
@@ -314,10 +322,12 @@ In order to support the above workflow the OONI API needs to support the followi
 
 In the following sections we will specify how these operations should be done.
 
-TODO(discuss): Do we want to support a delete operation? I would say we probably
-do not, since a user should always have access to the OONI Run link which they
-have configured on their device, if we do support deleting them, they might be
-surprised to see it disappear from their device in the future.
+By design, we don't specify a delete operation. This is because we want to
+ensure there is a permanent record of all OONI Run links that ever existed.
+
+We might however add support for rendering certain OONI Run links ineffective
+from the perspective of the application (i.e. making them not trigger any test
+runs).
 
 ## 4.1 CREATE a new OONI Run link
 
@@ -549,7 +559,9 @@ In particular the format of the OONI Run link should be such that if the URL is
 being truncated, it should be visible to the end, as such it's recommend that we
 restrict the character set of the `ooni_run_link_id` to just numbers. Since we
 might not end up having that many OONI Run link, this also lends itself well to
-allowing users to manually type OONI Run links directly into the app.
+allowing users to manually type OONI Run links directly into the app. When manually
+typing OONI Run links, the OONI Run link might be displayed broken up into
+numbers + spaces or dashes to make it easier to type.
 
 Mobile deep links can be registered using two different methods, one is a custom
 prefix (ex. `ooni://`), the other is a custom URL prefix (ex.
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -120,6 +120,8 @@ An OONI Run link descriptor is a JSON file with the following semantics:
 
   "author": "(optional) `string` name of the creator of this OONI Run link",
 
+  "is_archived": "(optional) `bool` a boolean flag used to indicate if this OONI Run link is archived. When an OONI Run link is archived, it does not run",
+
   // `array` provides a JSON array of tests to be run.
   "nettests":[{
 
@@ -325,9 +327,9 @@ In the following sections we will specify how these operations should be done.
 By design, we don't specify a delete operation. This is because we want to
 ensure there is a permanent record of all OONI Run links that ever existed.
 
-We might however add support for rendering certain OONI Run links ineffective
-from the perspective of the application (i.e. making them not trigger any test
-runs).
+A certain OONI Run link can rendered ineffective by setting the `is_archived`
+flag to true. In this case the OONI Run link still remains available, but it
+will not lead to tests being initiated.
 
 ## 4.1 CREATE a new OONI Run link
 
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -581,7 +581,9 @@ As such we recommend using the following addresses for OONI Run link and OONI Ru
 
 * `https://api.ooni.io/api/v1/ooni_run/{ooni_run_link_id}`
 
-Moreover, we could at some point host these links on s3 or github and have them
+# 6.0 Future work
+
+We could at some point host these links on s3 or github and have them
 be accessible via URLs in the form:
 
 * `https://raw.githubusercontent.com/ooni/run-links/master/data/{ooni_run_link_id}.json`
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -185,7 +185,7 @@ dashboard as follows.
          "options": {
             "MaxRuntime": "$settings.max_runtime",
          },
-         "test_settings": {
+         "backend_settings": {
             "category_codes": "$settings.category_codes"
          },
          "is_manual_run_enabled": true,
@@ -193,7 +193,7 @@ dashboard as follows.
          "test_name": "web_connectivity"
       },
       {
-         "test_settings": {
+         "backend_settings": {
             "category_codes": "$settings.category_codes"
          },
          "is_manual_run_enabled": false,
--- a/backends/bk-005-ooni-run.md
+++ b/backends/bk-005-ooni-run.md
@@ -1,6 +1,6 @@
 # OONI Run specification
 
-* author: Norbel Ambanumben
+* author: Norbel Ambanumben, Arturo Filastò
 * version: 0.0.1
 * date: 2022-06-15
 * status: _draft_
@@ -79,12 +79,221 @@ static, it should be possible server it from a mirror that provides higher
 levels of blocking resistance (and potentially some higher level of stealth),
 such as s3 or github.
 
-# 3.0 API
+# 3.0 OONI Run descriptor
+
+OONI Run descriptors define a set of experiments and configuration vectors used.
+It also includes additional metadata used to display a particular OONI Run object
+to the end user.
+
+Our goal is to use OONI Run descriptors to implement all the cards currently
+present inside of the OONI Probe app. As such we will be having a set of default
+OONI Run links that apps will ship with.
+
+An OONI Run link descriptor is a JSON file with the following semantics:
+```json
+{
+"name": "(required) `string` is the display name for the OONI Run link",
+// (optional) `map` of translations to language codes for the name
+"name_intl": {
+   "it": "Il nome del test in italiano",
+},
+// TODO: recommend a maximum length for this field
+"short_description": "(optional) `string` short_description for the OONI Run link.",
+// (optional) `map` of translations to language codes for the short_description
+"short_description_intl": {},
+"description": "(optional) `string` full description for this OONI Run link. This goes into the details of the card. Markdown is supported.",
+// (optional) `map` of translations to language codes for the description
+"description_intl": {
+   "it": "La descrizione del test in italiano"
+},
+"icon": "(optional) `string` the ID of any icon part of the OONI icon set",
+"author": "(optional) `string` name of the creator of this OONI Run link",
+// `array` provides a JSON array of tests to be run.
+"nettests":
+   [
+      {
+         // (optional) `array` provides a JSON array of tests to be run.
+         "inputs": [
+            "https://example.com/",
+            "https://ooni.org/"
+         ],
+         // (optional) `map` options arguments provided to the specified test_name
+         "options": {
+            "HTTP3Enabled": true,
+         },
+         // (optional) `map` settings which are sent to probe_services for retrieving the inputs vector
+         // It's possible to reference user configured settings by using accessing the 
+         // $settings special variable.
+         // In particular the content of the test_settings will be sent to the /api/v1/check-in call nested 
+         // under the relative test_name.
+         "test_settings": {
+            "category_codes": "$settings.category_codes"
+         },
+         // (optional) `bool` indicates if this test should be run as part of autoruns. Defaults to true.
+         "is_background_run_enabled": true,
+         // (optional) `bool` indicates if this test should be run as part of manual runs. Defaults to true.
+         "is_manual_run_enabled": true,
+         "test_name": "web_connectivity"
+      },
+      {
+         "test_name": "openvpn",
+         "test_settings": {
+            "provider": "riseupvpn"
+         },
+      }
+   ],
+}
+```
+
+Based on the above specification it would be possible to re-implement the cards for the OONI Probe
+dashboard as follows:
+
+### Websites
+
+```json
+{
+"name": "Websites",
+"short_description": "Test the blocking of websites",
+"description": "Check whether websites are blocked using OONI's [Web Connectivity test](https://ooni.org/nettest/web-connectivity/)...",
+"icon": "OONINettestGroupWebsites",
+"author": "contact@ooni.org",
+"nettests":
+   [
+      {
+         "options": {
+            "MaxRuntime": "$settings.max_runtime",
+         },
+         "test_settings": {
+            "category_codes": "$settings.category_codes"
+         },
+         "is_manual_run_enabled": true,
+         "is_background_run_enabled": false,
+         "test_name": "web_connectivity"
+      },
+      {
+         "test_settings": {
+            "category_codes": "$settings.category_codes"
+         },
+         "is_manual_run_enabled": false,
+         "is_background_run_enabled": true,
+         "test_name": "web_connectivity"
+      },
+   ]
+}
+```
+
+### Instant Messaging
+
+```json
+{
+"name": "Instant Messaging",
+"short_description": "Test the blocking of instant messaging apps",
+"description": "Check whether [WhatsApp](https://ooni.org/nettest/whatsapp/), ...",
+"icon": "OONINettestGroupInstantMessaging",
+"author": "contact@ooni.org",
+"nettests":
+   [
+      {
+         "test_name": "whatsapp"
+      },
+      {
+         "test_name": "telegram"
+      },
+      {
+         "test_name": "facebook_messenger"
+      },
+      {
+         "test_name": "signal"
+      },
+   ]
+}
+```
+
+### Instant Messaging
+
+```json
+{
+"name": "Circumvention",
+"short_description": "Test the blocking of censorship circumvention tools",
+"description": "Check whether [Psiphon](https://ooni.org/nettest/psiphon/), ...",
+"icon": "OONINettestGroupCircumvention",
+"author": "contact@ooni.org",
+"nettests":
+   [
+      {
+         "test_name": "psiphon"
+      },
+      {
+         "test_name": "tor"
+      },
+      {
+         "test_name": "riseupvpn"
+      }
+   ]
+}
+```
+
+### Performance
+
+```json
+{
+"name": "Performance",
+"short_description": "Test your network speed and performance",
+"description": "Measure the speed and performance of your network using the [NDT](https://ooni.org/nettest/ndt/) test. ...",
+"icon": "OONINettestGroupPerformance",
+"author": "contact@ooni.org",
+"nettests":
+   [
+      {
+         "test_name": "http_invalid_request_line"
+      },
+      {
+         "test_name": "http_header_field_manipulation"
+      },
+      {
+         "test_name": "ndt"
+      },
+      {
+         "test_name": "dash"
+      },
+   ]
+}
+```
+
+### Experimental
+
+```json
+{
+"name": "Experimental",
+"short_description": "Run new experimental tests",
+"icon": "OONINettestGroupExperimental",
+"author": "contact@ooni.org",
+"nettests":
+   [
+      {
+         "test_name": "stun_reachability"
+      },
+      {
+         "test_name": "dnscheck"
+      },
+      {
+         "is_manual_run_enabled": false,
+         "test_name": "tor_snowflake"
+      },
+      {
+         "is_manual_run_enabled": false,
+         "test_name": "vanilla_tor"
+      },
+   ]
+}
+```
+
+# 4.0 API
 
 In order to support the above workflow the OONI API needs to support the following operations:
-* CREATE a new OONI Run link, returning the OONI Run link ID (see 3.1)
-* UPDATE an existing OONI Run link (see 3.2)
-* GET the OONI Run descriptor, provided an ID (3.3)
+* CREATE a new OONI Run link, returning the OONI Run link ID (see 4.1)
+* UPDATE an existing OONI Run link (see 4.2)
+* GET the OONI Run descriptor, provided an ID (4.3)
 
 In the following sections we will specify how these operations should be done.
 
@@ -93,7 +302,7 @@ do not, since a user should always have access to the OONI Run link which they
 have configured on their device, if we do support deleting them, they might be
 surprised to see it disappear from their device in the future.
 
-## 3.1 CREATE a new OONI Run link
+## 4.1 CREATE a new OONI Run link
 
 This operation will be performed by a logged in user that is interested in
 performing an OONI Run link based measurement campaign.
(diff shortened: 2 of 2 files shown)


backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
+# OONI Run specification
+
+* author: Norbel Ambanumben
+* version: 0.0.1
+* date: 2022-06-15
+* status: _draft_
+
+This document aims at providing a functional specification for OONI Run.
```
bassosimone, line 8:
```suggestion
This document provides a functional specification for OONI Run.
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+# 1.0 System overview
+
+The api exposes an HTTP API allowing OONI Run web app to submit links to the
```
bassosimone, line 12:
```suggestion
The ooni/api exposes an HTTP API allowing OONI Run web app to submit links to the
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+us or be prepared to retry without compression.
+
+New implementations MUST properly set `Content-Type`. Server side
+implementatons MUST be able to deal with legacy clients that possibly
+do not correctly set the `Content-Type`.
+
+Tests submitted to the OONI Run API will be stored and processed by OONI backend
+and OONNI Probes the URL is shared with.
+How that will happen is out of the scope of this document. You may also want to read
+[about our data policy](https://ooni.io/about/data-policy/).
```
bassosimone, line 30:
```suggestion
Tests submitted to the OONI Run API will be stored and processed by OONI backend
and OONI Probes the URL is shared with.
How that will happen is out of the scope of this document. You may also want to read
[about our data policy](https://ooni.io/about/data-policy/).
```

That said, I think this paragraph is not very relevant for this specification.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+The OONI Run MUST be exposed as an HTTPS service. It MUST also be exposed as
+a Tor onion service as long as legacy OONI probe clients use it. The need
```
bassosimone, line 33: There are no legacy OONI Probe clients using OONI Run. So, this paragraph should probably be simplified.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+to expose an onion service will be rediscussed when legacy OONI probe
+clients will no longer be relevant. A [legacy document](
+https://ooni.torproject.org/docs/architecture.html)
+explains why the OONI project originally chose to allow for both HTTPS
+and Tor onion service services (henceforth, Onion).
+
+It is also outside of the scope of this section to define the way in which
+a OONI probe discovers the collector API endpoint, as well as how, given
+several endpoints and/or endpoint types (HTTPS, Onion), it chooses a specific
+endpoint.
```
bassosimone, line 43: Please remove this paragraph. It's about the collector.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+# 2.0 Threat model
+
+The OONI RUN Service transport MUST guarantee some reasonable level of encryption
+and authentication between the OONI probe and itself. Therefore, a malicious
+lazy actor won't be able to easily modify test results that they have not
+created while they are being submitted to the OONI Run Service. Also, they will not
+be able to easily log the messages exchanged with the OONI Run Service.
```
bassosimone, line 51: Please change the wording. This wording clearly still refers to the collector.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+- you access it from probes or frontend service (see 3.3);
+
+- then, you can disable/delete it (see 3.4).
+
+## 3.1 Create a new OONI Run link
+
+### Request
+
+When a you *create* a new OONI RUN link, the frontend sends a `POST`
```
bassosimone, line 79:
```suggestion
When you *create* a new OONI RUN link, the frontend sends a `POST`
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+        (required) `string` it is required property used in identifying the link in the OONI probe and other parts of the system.
+
+     "description":
+        (optional) `string` describing the tests included as part of the OONI Run link.
+      
+      "tests": `array` provides a JSON array of tests to be run.
+         [
+            {
+                  "ta": {
+                     "urls": []
```
bassosimone, line 95:
```suggestion
                     "inputs": []
```

We're already cursed enough by the fact that we think the input the experiments are URLs. Let's be more generic and allow more flexibility. It will pay off in the future.

(I know the v1 format says `urls` but this is v2 and we should make it better, so...)

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+compliant, serialized JSON object.
+
+### Response status code
+
+Upon receiving a request to create a report, the OONI Run Service:
+
+1. SHOULD fail with `4xx` if the request body does not parse, it is not a JSON object,
+   any required field is missing and/or if any present field has an invalid value.
+
+2. MUST fail with `4xx` if the request is not compliant with its policies.
```
bassosimone, line 118: Which policies?

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+        specified, except that is MUST be a valid UTF-8 string, of
+        course. The client MUST NOT make any assumption with respect
+        to the structure of this field.
+    }
+
+### Response processing requirements
+
+Upon receiving a response, new written clients MUST check that the
+status is `200` before continuing, MUST ensure that the server supports
+their preferred data submission format, and MUST save the OONI Run ID.
```
bassosimone, line 150: Please edit to remove unnecessary wording borrowed from the collector spec

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+## 3.2 Update an OONI Run Link
+
+Updating an OONI Run Link means appending a test to the OONI Run Link.
```
bassosimone, line 154: What about adding new URLs?

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+            },
+            {
+                  "tn": "signal"
+            }
+         ]
+    }
+
+### Response status code
+
+Upon receiving this request, the collector:
```
bassosimone, line 188:
```suggestion
Upon receiving this request, the OONI Run backend:
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+3. MUST "commit" the OONI Run data to persistent storage or to some
+   database before returning `200` to the client making
+   sure that it successfully saved the OONI Run Data (e.g. by
+   checking the return value of `fclose`). There SHOULD be
+   integration tests to check whether the JSON serialized by
+   a specific implementation is compliant with the format
+   expected by the pipeline (e.g. in `go` with `omitempty`
+   `null` values are removed by a marshal-serialize cycle
+   while the Python parser preserves them).
```
bassosimone, line 204: I don't think we need to be so strict. This wording was very careful because we were describing measurement submission but I don't think there are the same requirements here.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+4. is allowed to perform additional quick operations that may have
+   an impact on the status code, and SHOULD defer other operations
+   that do not have an impact on the status code to after the status
+   code has been sent to the client (the goal being to keep the
+   connection open for as little as possible to avoid the risk that
+   a middlebox in a constrained network flags the connection as
+   stale and closse it, thus preventing the server from telling the client
+   that the measurement has been successfully submitted).
```
bassosimone, line 213: Not sure if this is needed here

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+return a `{"error": "string"}` object to make debugging easier.
+
+In case of `200` responses, new OONI Run Service implementations MAY
+include a unique identifier for the OONI Run Link that the
+client may later use to reference said OONI Run Link. For example:
+
+
+```JSON
+{"ooni_run_id":"e00c584e6e9e5326"}
+```
```
bassosimone, line 230: I'm not sure about returning the ID here. We already know it because we just modified the OONI Run link.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+- then, you can disable/delete it (see 3.4).
+
+## 3.1 Create a new OONI Run link
+
+### Request
+
+When a you *create* a new OONI RUN link, the frontend sends a `POST`
+request conforming to the following, informal specification:
+
+    POST /link
```
bassosimone, line 82:
```suggestion
    POST /api/v1/oonirun/link
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+## 3.2 Update an OONI Run Link
+
+Updating an OONI Run Link means appending a test to the OONI Run Link.
+
+### Request
+
+To update an OONI Run Link, the client issues a request compliant with:
+
+    POST /report/${ooni_run_id}
```
bassosimone, line 160:
```suggestion
    POST /api/v1/oonirun/link/${ooni_run_id}
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+As far as the client is concerned, this `ooni_run_id` is an
+opaque UTF-8 string that has some meaning to the server.
+
+## 3.3 Deleting an OONI Run Link
+
+To delete a report, a client should submit a request like:
+
+    DELETE /link/${ooni_run_id}
```
bassosimone, line 239:
```suggestion
    DELETE /api/v1/oonirun/link/${ooni_run_id}
```

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+If the report exists, `200` is returned
+and the response MUST include this body for backwards
+compatibility with existing implementations:
+
+```JSON
+{"status": "success"}
+```
+
+Otherwise, in case of failure, the OONI Run Service MUST return a JSON object, whose
+content is implementation dependent and MAY be empty. Some existing implementations
```
bassosimone, line 250: Please mention that it should return `4xx` on failure

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+return a `{"error": "string"}` object to make debugging easier.
+
+# 4.0 Implementation considerations
+
+A client side implementation of the OONI Run Service MUST make sure
+that it is emitting timestamps using UTC rather than local time.
```
bassosimone, line 256: Not sure this is relevant

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+A client side implementation MAY retry any failing OONI Run Service operation
+immediately for three times in case there is a DNS or TCP error. This
+is to ensure that transient errors do not prevent us from submitting the
+data immediately. If all these immediate retries fail, then the client
+SHOULD arrange for resubmitting the OONI Run Link at a later time, either
+requiring user input or automatically. In the latter case, the delay
+after which the client will attempt to resubmit SHOULD be exponentially
+distributed and SHOULD NOT be smaller than 15 minutes.
```
bassosimone, line 265: Not sure this is relevant

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+A server implementation MAY publish metrics allowing OONI to gradually
+enforce more JSON schema correctness in the OONI Run Service, both for the
+open-report request and for submitted data. A possible strategy to make
+this happen consists of parsing without rejecting, counting the number
+of failures, and then iterating until the number of failures starts
+converging to zero. At that point, it is possible to make the checks
+mandatory and reject invalid input.
```
bassosimone, line 273: Not sure this is relevant

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+        (required) `string` it is required property used in identifying the link in the OONI probe and other parts of the system.
+
+     "description":
+        (optional) `string` describing the tests included as part of the OONI Run link.
+      
+      "tests": `array` provides a JSON array of tests to be run.
+         [
+            {
+                  "ta": {
+                     "urls": []
```
bassosimone, line 173:
```suggestion
                     "inputs": []
```

Same rationale as above

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+is implementation dependent and MAY be empty. Some existing implementations
+return a `{"error": "string"}` object to make debugging easier.
+
+In case of success (i.e. `200` response), the OONI Run Service MUST return a JSON body
+generated in compliance with the following, informal specification:
+
+    {
+      "backend_version":
+        `string` containing the version of the backend. This version MUST
+        match the regex provided above for 'software_version'.
```
bassosimone, line 137: I am not a fan of returning this information. We used to do that in the collect for legacy reasons. I think we may also decide to avoid returning this information, actually.

bassosimone (commented):
It's a good starting point. Please, apply the requested changes or discuss why you think they should not be implemented if you disagree with me. Thank you! 💯

bassosimone:
Forgot to mention: we should also specify how a canonical OONI Run v2 link looks like. Because we're using deep linking, I assume it could be something like `https://run.ooni.io/api/v2/nettest/ID` or `ooni://nettest/v2/ID`.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+   `null` values are removed by a marshal-serialize cycle
+   while the Python parser preserves them).
+
+4. is allowed to perform additional quick operations that may have
+   an impact on the status code, and SHOULD defer other operations
+   that do not have an impact on the status code to after the status
+   code has been sent to the client (the goal being to keep the
+   connection open for as little as possible to avoid the risk that
+   a middlebox in a constrained network flags the connection as
+   stale and closse it, thus preventing the server from telling the client
```
bassosimone, line 212:
```suggestion
   stale and close it, thus preventing the server from telling the client
```

ainghazal:
Since conversation for this spec seems to be still open, I'll leave here a comment that I made recently to @bassosimone: if the limits in the information that can be conveyed in the url params is the limiting factor to move to a backend-based solution, we could consider encoding the information in some other way?

One obvious candidate is QRs. They're quite ubiquitous nowadays, and registering the right intent they might perhaps be able to encode a complete experiment run (we can try to pack it more efficiently; and perhaps a little research into information density of different types of QR is in order.

This is not to say not to pursue a backend (a backend with community control brings a lot of advantages), but perhaps some runs do not need of it after all.

hellais:
I like the QR code idea! However I am not sure how practical it's going to be. In practice the use-case for sharing OONI Run links is that people share them through digital means (ex. IM or social media). If you are to use a QR code image on social media, it's going to be tricky to scan the QR code from social media or IM unless you have two devices.
Unless there is some other way to go from a QR code inside of say WhatsApp to a the native phone app for handing the code.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+This document provides a functional specification for OONI Run.
+
+# 1.0 System overview
+
+The ooni/api exposes an HTTP API allowing OONI Run web app to submit links to the
+backend and also allow the OONI Probe to query OONI Run link data.
+The client and the server MUST NOT assume a keep
+alive semantics for the HTTP connections.
+
```
hellais, line 16: I think we should add a list of definitions for what we mean when we talk about "OONI Run link" and "OONI Run link data".

I suggest defining the following terms:
- OONI Run link - This is the mobile deep link that when a user taps on it having the OONI Probe app installed, will trigger opening the app and be handled by it. If it's accessed from a device that doesn't have the app installed (or the handling is broken) it will display a web page prompting the user to install the app or allowing them to force opening the app with the custom scheme `ooni://`
- OONI Run descriptor - This is the metadata for an experiment run
- OONI Run descriptor URL - This is the URL used to retrieve the descriptor from a backend

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+New server implementations MUST support `Content-Encoding: gzip` to
+compress request bodies. OONI production clients MUST NOT use this
+feature until all OONI's production servers server are known to have
+implemented it. Third party clients should either coordinate with
+us or be prepared to retry without compression.
```
hellais, line 21: This is not relevant to this OONI Run.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+New implementations MUST properly set `Content-Type`. Server side
+implementatons MUST be able to deal with legacy clients that possibly
+do not correctly set the `Content-Type`.
```
hellais, line 25: This is not relevant to OONI Run. There is no notion of "New" implementations since we never had a backend for OONI Run.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+clients will no longer be relevant. A [legacy document](
+https://ooni.torproject.org/docs/architecture.html)
+explains why the OONI project originally chose to allow for both HTTPS
+and Tor onion service services (henceforth, Onion).
+
+It is also outside of the scope of this section to define the way in which
+a OONI probe discovers the collector API endpoint, as well as how, given
+several endpoints and/or endpoint types (HTTPS, Onion), it chooses a specific
+endpoint.
+
```
hellais, line 44: I think more details need to be provided on the high level architecture of how this specific system is going to work. Specifically we need to document how we go from handling an OONI Run link to retrieving the OONI Run descriptor via the OONI Run descriptor URL

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+to conceal to a passive network observer the fact that they are communicating to
+a OONI Run Service. Such properties are to be provided by other software, e.g. Tor.
+
+Therefore a client implementation of the OONI Run Service protocol SHOULD allow one
+to specify a [SOCKS5](https://tools.ietf.org/html/rfc1928) proxy where the
+name resolution is performed by the circumvention tool (`socks5h`).
+
+# 3.0 API
+
+The same API is available via HTTP and Onion.
```
hellais, line 63: I don't think we should be exposing this via onion service.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+
+The standard flow for OONI RUN Link is the following:
+
+- you create an OONI RUN Link (see 3.1);
```
hellais, line 67: I think it's important here to clarify the various components you are interacting with when creating an OONI Run link and what are the bits of information that you get back when creating the link.

For example at some point you are interacting with a web service that speak to an API backend and the thing you get back is an OONI Run link that when opening with the mobile app speaks to the API backend to retrieve the OONI Run descriptor.

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,273 @@
...
+- then, you can disable/delete it (see 3.4).
+
+## 3.1 Create a new OONI Run link
+
+### Request
+
+When you *create* a new OONI RUN link, the frontend sends a `POST`
+request conforming to the following, informal specification:
+
+    POST /api/v2/oonirun/link
```
hellais, line 82: I think this should still be nested under `/api/v1/` since the versioning is of the OONI API not the OONI Run link itself. As we don't have a `v1` OONI Run link speaking to the API we can use this without any issues.

I suggest we use as an endpoint:
```
/api/v2/oonirun/link
```

hellais:
Following a discussion I had with @ainghazal, I have updated the spec to include more details about the descriptor format.

I also factored into it the plans we had for redoing the dashboard based on the OONI Run v2 format and provided examples for how we could implement the dashboard using them.

To show how this could also work for the use-case of @ainghazal, we could imagine having an OONI Run v2 link for let's say RiseupVPN which could look like this:

```
{
"name": "RiseupVPN",
"short_description": "Run experiments related to testing RiseupVPN",
"author": "contact@ooni.org",
"nettests":
   [
      {
         "inputs": [
           "https://api.black.riseup.net:443/3/config/eip-service.json",
           "https://riseup.net/provider.json",
           "https://api.black.riseup.net:9001/json",
         ],
         "options": {
           "SkipTLSVerify": true
         },
         "test_name": "web_connectivity"
      },
      {
         "test_settings": {
            "provider": "riseupvpn"
          },
         "test_name": "openvpn"
      },
   ]
}
``` 

Where `SkipTLSVerify` should be some options understood by the OONI Probe engine to skip certificate verification (we need this since the riseup endpoints are using their own root CA) and `provider` is expected to be understood by the `check-in` API to derive the input and configuration vectors for the `openvpn` test.

We could take this approach even further and consider applying a similar treatment to all the IM tests and rebuild them using these more basic building blocks.

This does however raise the question of whether or not we would like to have an additional layer of grouping/nesting for the nettests.

For example I might want to put the RiseupVPN OONI Run link inside of the Circumvention test group, but as it's done now, I am not able to do so because the structure is too flat.

We could do this in at least two different ways.

1. We just add a tag inside of each nettest, that the apps will then know to group by
ex.
```json
{
"name": "VPN tests",
"short_description": "Run experiments related to testing VPNs",
"author": "contact@ooni.org",
"nettests":
   [
      {
         "inputs": [
           "https://api.black.riseup.net:443/3/config/eip-service.json",
           "https://riseup.net/provider.json",
           "https://api.black.riseup.net:9001/json",
         ],
         "options": {
           "SkipTLSVerify": true
         },
         "tag": "riseupvpn",
         "test_name": "web_connectivity"
      },
      {
         "test_settings": {
            "provider": "riseupvpn"
          },
         "tag": "riseupvpn",
         "test_name": "openvpn"
      },
      {
         "inputs": [
           "https://tunnebear.com/targets.json",
         ],
         "tag": "tunnelbear",
         "test_name": "web_connectivity"
      },
      {
         "test_settings": {
            "provider": "riseupvpn"
          },
         "tag": "tunnelbear",
         "test_name": "openvpn"
      },
   ]
}
```

2. We add an extra level of nesting in the nettests

1. is probably easier and it doesn't require changing the data format, but might limit our ability to enrich the tags with extra metadata (like title and description related to a certain tag).

Xredmi:
> ## Description
> Spec for OONI Run Service

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,576 @@
...
+allowing users to manually type OONI Run links directly into the app.
+
+Mobile deep links can be registered using two different methods, one is a custom
+prefix (ex. `ooni://`), the other is a custom URL prefix (ex.
+`https://run.ooni.io/1234`). In our testing we have seen that the custom prefix
+is more reliable, yet it has the tradeoff of not allowing us to display a web
+page when the user does not have the app installed. As such the recommended
+strategy is to encourage users to share the custom URL prefix OONI Run link, but
+on the web page itself, in the event that the app did not handle the deep link,
+have a link to the custom prefix approach to "force" the opening of the app.
```
bassosimone, line 561: I think we should clarify the last sentence. Does it mean that the user will have a webpage with a link and the _option_ of clicking on the link to trigger the app?

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,576 @@
...
+As such we recommend using the following addresses for OONI Run link and OONI Run descriptor URLs:
+
+* `https://run.ooni.io/{ooni_run_link_id}`, where `{ooni_run_link_id}` is a number
+
+* `ooni://runv2/{ooni_run_link_id}`
+
+* `https://api.ooni.io/api/v1/ooni_run/{ooni_run_link_id}`
+
+Moreover, we could at some point host these links on s3 or github and have them
+be accessible via URLs in the form:
```
bassosimone, line 572: I think it is unclear how authentication and installation would work in this case. I suggest we move this last portion of the document into its own section, named "Future work".

backends/bk-005-ooni-run.md
```diff
@@ -0,0 +1,576 @@
+# OONI Run specification
+
+* author: Norbel Ambanumben, Arturo Filastò
+* version: 2023.02.09
```
bassosimone, line 4: I removed version. We already have date and we can format dates to be version numbers, like I did. It is already burdensome to maintain one identifier, so we should not maintain two. We should migrate all specs to have their specification version number being the date of the last modification because there is no point in saying "I am implementing version 0.0.1 of oonirun v2 specification. We always implement the last version and we already factor in breaking changes by changing the name of the spec. For example, `ooni-run` and `ooni-run-v2`. In this vein, I am going to clearly name the file such that it's obvious we're referring to `ooni-run-v2`.

backends/bk-005-ooni-run-v2.md
```diff
@@ -0,0 +1,576 @@
...
+    "options": {
+      "HTTP3Enabled": true
+    },
+
+    // (optional) `map` settings which are sent to probe_services for retrieving the inputs vector
+    // It's possible to reference user configured settings by using accessing the 
+    // $settings special variable.
+    // In particular the content of the test_settings will be sent to the /api/v1/check-in call nested 
+    // under the relative test_name.
+    "test_settings": {
```
bassosimone, line 142: This field cannot be named `test_settings` because it makes super confusing for anyone who is not super familiar with the spec to read a JSON and do not know which are options for the experiment and which are options for the backend. We should rename this field `backend_options`.

backends/bk-005-ooni-run-v2.md
```diff
@@ -0,0 +1,576 @@
...
+  
+    // (optional) `array` provides a JSON array of tests to be run.
+    "inputs": [
+      "https://example.com/",
+      "https://ooni.org/"
+    ],
+    
+    // (optional) `map` options arguments provided to the specified test_name
+    "options": {
+      "HTTP3Enabled": true
```
bassosimone, line 134: We have a `map[string]any` data format here. This is great. However, the base data format for experiments says that options are `[]string` instead. We need to patch the way in which we include options into the base data format to be a `map[string]any`. We should do this by adding a new field, to avoid breaking backwards compatibility (assuming we - or anyone - is parsing the options).

backends/bk-005-ooni-run-v2.md
```diff
@@ -0,0 +1,576 @@
...
+* UPDATE an existing OONI Run link (see 4.2)
+
+* GET the OONI Run descriptor, provided an ID (4.3)
+
+In the following sections we will specify how these operations should be done.
+
+TODO(discuss): Do we want to support a delete operation? I would say we probably
+do not, since a user should always have access to the OONI Run link which they
+have configured on their device, if we do support deleting them, they might be
+surprised to see it disappear from their device in the future.
```
bassosimone, line 320: We will for sure want to delete some descriptors _in some cases_. I think the text should say that we will not implement deletion as a _removal_ of the descriptor but we will allow parking a descriptor in such a way that it has no effect on the app in terms of running tests. In other words, I think that deleting the descriptor is more a social than a technical need and it should be enough to put the descriptor into a "null" state to achieve the social needs without breaking the installed applications too much.

backends/bk-005-ooni-run-v2.md
```diff
@@ -0,0 +1,576 @@
...
+      "HTTP3Enabled": true
+    },
+
+    // (optional) `map` settings which are sent to probe_services for retrieving the inputs vector
+    // It's possible to reference user configured settings by using accessing the 
+    // $settings special variable.
+    // In particular the content of the test_settings will be sent to the /api/v1/check-in call nested 
+    // under the relative test_name.
+    "test_settings": {
+      "category_codes": "$settings.category_codes"
```
bassosimone, line 143: I think this spec should do a better job at explaining what `$settings` is and means. I also think it may be worth considering checking whether we could reimplement the "decks" using this format: the "decks" where quite a flexible and nice format, so if we can implement the same functionality with OONI Run v2, we are definitely well positioned with this work in terms of flexibility.

backends/bk-005-ooni-run-v2.md
```diff
@@ -0,0 +1,576 @@
...
+}
+```
+
+# 5.0 Implementation considerations
+
+Special attention should be placed in ensuring the OONI Run links (which are
+mobile deep links) are sharable though various apps.
+In particular the format of the OONI Run link should be such that if the URL is
+being truncated, it should be visible to the end, as such it's recommend that we
+restrict the character set of the `ooni_run_link_id` to just numbers. Since we
```
bassosimone, line 550: I disagree with "just" numbers. I think we should figure out a mechanism by which the real ID is just numbers but we can express the ID using `{4-digit} "-" {4-digit} ...`. I say this because, if we want to allow users to manually insert OONI Run IDs, then we need to make it simpler for that to split the digits.

bassosimone (changes requested):
To summarize my request for changes:

- [ ] we need to rename `test_settings` to `backend_options`  because the original naming for the field allows for too much confusion between it and the `options

- [ ] we need to agree (or discuss further) the need to have a batter definition for options inside the base data format

- [ ] it is unclear how we're going to allow for CRUD operations for URLs deployed elsewhere, so we should be careful with the language and frame this as future work

- [ ] we need to specify a lightweight socially acceptable mean of deleting a OONI Run URL that perhaps does not vanish the URL but parks it forever as "gone"

- [ ] we should explain what `$settings` means in the spec and how in practice the app is going to process this kind of fields

- [ ] we should probably cross check with the old "decks" format and make sure that we can support a similar level of complexity or justify why this design is simpler (in many cases simpler is better, so I am not saying we should support all the "decks" features necessarily but that we should be asking this question)

- [ ] if adding OONI Run IDs manually is really a goal, then I suggest we figure out a way by which the link can contain numbers and dashes, to help a human to split the digits apart and avoid the risk that someone types in the wrong ID. (It may also be useful have some digits be some form of checksum for other digits, otherwise, to have more robustness).

backends/bk-005-ooni-run-v2.md
hellais: Yes I will add a section explaining these.

backends/bk-005-ooni-run-v2.md
hellais: Added a sentence on this.

backends/bk-005-ooni-run-v2.md
hellais: Yes, I think that is a matter of presenting the input UI in a way that is conveyed, but we can still keep the link itself limited to just numbers.

I think having dashes in the links runs the risk of users not noticing that link is truncated (for example if they are clicking on it in an app which may truncate links with dashes).

There is a difference between what is the content of the link itself vs how a user might manually enter the link into the field where they can manually enter it.

An example of this is a credit card number. The number itself doesn't contain spaces, yet it's common for user interfaces to break up the number into blocks of numbers and spaces. We can do the same thing with OONI Run links.

hellais: I added a sentence to further clarify that.

backends/bk-005-ooni-run.md
hellais: Yes that's correct. Do you have a suggestion on how to improve the text? I added a sentence at the end saying this is how OONI Run links currently work.

backends/bk-005-ooni-run.md
hellais: Moved

backends/bk-005-ooni-run-v2.md
hellais: Yes this makes sense. I don't think we need to edit anything in the spec though, correct?

bassosimone: correct!

bassosimone (approved):
🐳

mScott224 (approved):

PR #255, opened 2022-08-10
Openvpn experiment spec

ainghazal:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2202
- [x] related ooni/probe-cli pull request: https://github.com/ooni/probe-cli/pull/845

Location of the issue tracker: https://github.com/ooni/probe

## Description

This is a first draft of the openvpn experiment spec, in its current iteration.

### DIFF
--- a/data-formats/df-001-httpt.md
+++ b/data-formats/df-001-httpt.md
@@ -122,6 +122,10 @@ are multiple values for the same header key. See below the definition of
 - `tor` (`TorInfo`): this is an object containing information on the
 instance of `tor` that we may be using for measuring.
 
+- `vpn` (`VPNInfo`): similarly to `tor`, this is an object containing
+  information on any `vpn` instance that we may be using for
+  measuring.
+
 - `x_transport` (`string`; deprecated): indicates what transports was used for
 issuing the request. Typically `"tcp"` or `"udp"`. Until 2022-09-08, we used
 `"tcp"` or `"quic"` but we realized that using `"udp"` instead of `"quic"` would
@@ -222,6 +226,24 @@ if we're not using tor or this information is not available.
 if we're not using tor or this information is not available.
 
 - `is_tor` (`bool`): true if we're using tor, false otherwise.
+ 
+## VPNInfo
+
+```JSON
+{
+    "exit_ip": null,
+    "provider": null,
+    "is_vpn": false
+}
+
+- `exit_ip` (`string`; nullable): the egress IP for the VPN gateway, or `null`
+  if we're not using a VPN or this information is not available.
+
+- `provider` (`string`; nullable): the name of the provider to which this VPN
+  gateway belongs to, or `null` if we're not using a VPN or this information is
+  not available.
+
+- `is_vpn` (`bool`): true if we're using a VPN, false otherwise.
 
 ## Example
 
@@ -271,6 +293,11 @@ not relevant to the HTTP data format:
       "exit_ip": null,
       "exit_name": null,
       "is_tor": false
+    },
+    "vpn": {
+      "exit_ip": null,
+      "provider": null,
+      "is_vpn": false
     },
     "x_transport": "tcp",
     "url": "https://example.com"
--- a/data-formats/df-001-httpt.md
+++ b/data-formats/df-001-httpt.md
@@ -122,9 +122,8 @@ are multiple values for the same header key. See below the definition of
 - `tor` (`TorInfo`): this is an object containing information on the
 instance of `tor` that we may be using for measuring.
 
-- `vpn` (`VPNInfo`): similarly to `tor`, this is an object containing
-  information on any `vpn` instance that we may be using for
-  measuring.
+- `vpn` (`VPNInfo`; optional): similarly to `tor`, this is an object containing 
+information on any `vpn` instance that we may be using for measuring.
 
 - `x_transport` (`string`; deprecated): indicates what transports was used for
 issuing the request. Typically `"tcp"` or `"udp"`. Until 2022-09-08, we used
@@ -232,19 +231,21 @@ if we're not using tor or this information is not available.
 ```JSON
 {
     "exit_ip": null,
-    "provider": null,
-    "is_vpn": false
+    "gateway": null,
+    "provider": null
 }
 ```
 
 - `exit_ip` (`string`; nullable): the egress IP for the VPN gateway, or `null`
-  if we're not using a VPN or this information is not available.
+if we're not using a VPN or this information is not available.
 
-- `provider` (`string`; nullable): the name of the provider to which this VPN
-  gateway belongs to, or `null` if we're not using a VPN or this information is
-  not available.
+- `gateway` (`string`; nullable): an identifier for the gateway to where the tunnel 
+is connecting to. It's up to the implementor of the experiment to decide whether this 
+identifier is an IP address, or some other form of possibly anonymized
+identification.
 
-- `is_vpn` (`bool`): true if we're using a VPN, false otherwise.
+- `provider` (`string`; nullable): the name of the provider to which this VPN 
+gateway belongs to.
 
 ## Example
 
--- a/data-formats/df-001-httpt.md
+++ b/data-formats/df-001-httpt.md
@@ -235,6 +235,7 @@ if we're not using tor or this information is not available.
     "provider": null,
     "is_vpn": false
 }
+```
 
 - `exit_ip` (`string`; nullable): the egress IP for the VPN gateway, or `null`
   if we're not using a VPN or this information is not available.
--- a/nettests/ts-039-openvpn.md
+++ b/nettests/ts-039-openvpn.md
@@ -25,7 +25,7 @@ options.
 Altogether, they constitute a basic subset of the valid configuration
 parameters for the reference openvpn implementation (`man 8 openvpn`).
 
-The `URI` encodes information for the protocol (`openvpn`, in this case),
+The input `URI` encodes information for the protocol (`openvpn`, in this case),
 provider, transport, obfuscation and the remote address.
 
 ```
@@ -48,6 +48,12 @@ Other valid options, that will be passed to the openvpn invocation, are:
 
 - `cipher` (one of: `AES-256-GCM`, `AES-128-GCM`, `AES-256-CBC`, `AES-128-CBC`).
 - `auth` (one of: `SHA1`, `SHA512`).
+
+Additionally, the user can specify any amount of comma-separated urls that
+wants to be fetched through the tunnel. 
+
+- `URLs` (`string`, optional).
+
 
 # Test description
 
--- a/nettests/ts-039-openvpn.md
+++ b/nettests/ts-039-openvpn.md
@@ -113,137 +113,95 @@ abbreviated for clarity):
 
 ```JSON
 {
-  "test_keys": {
     "provider": "riseup",
     "vpn_protocol": "openvpn",
     "transport": "tcp",
-    "remote": "51.158.144.32:80",
-    "obfuscation": "none",
-    "bootstrap_time": 0.37221659,
-    "minivpn_version": "(devel)",
-    "obfs4_version": "",
-    "success": true
-    "stages": [
+    "remote": "198.252.153.109:443",
+    "obfuscation": "obfs4",
+    "bootstrap_time": 1.5671859910000001,
+    "network_events": [
       {
-        "op_id": 0
+        "transaction_id": 0,
         "operation": "ready",
-        "t": 0.145
+        "t": 0.12
       },
       {
-        "op_id": 8,
+        "transaction_id": 1,
+        "operation": "dial_done",
+        "t": 515.117
+      },
+      {
+        "transaction_id": 2,
+        "operation": "vpn_handshake_start",
+        "t": 515.118
+      },
+      {
+        "transaction_id": 3,
+        "operation": "reset",
+        "t": 515.119
+      },
+      {
+        "transaction_id": 5,
+        "operation": "tls_handshake_start",
+        "t": 771.98
+      },
+      {
+        "transaction_id": 6,
+        "operation": "tls_handshake_done",
+        "t": 1041.248
+      },
+      {
+        "transaction_id": 7,
+        "operation": "data_init",
+        "t": 1567.111
+      },
+      {
+        "transaction_id": 8,
         "operation": "vpn_handshake_done",
-        "t": 372.14
+        "t": 1567.114
       }
     ],
+    "last_handshake_transaction_id": 8,
     "tcp_connect": {
-      "ip": "51.158.144.32",
-      "port": 80,
+      "ip": "198.252.153.109",
+      "port": 443,
       "status": {
         "failure": null,
         "success": true
       },
-      "t0": 0.000239501,
-      "t": 0.045310671,
-      "transaction_id": 1
+      "t0": 0.000424974,
+      "t": 0.258504018
     },
     "failure": null,
-    "error": null,
     "icmp_pings": [
       {
         "target": "8.8.8.8",
         "sequence": [
           {
-            "seq": 9,
-            "ttl": 113,
-            "rtt": 60.208
-          },
-          {
-            "seq": 10,
-            "ttl": 113,
-            "rtt": 57.909
+            "seq": 1,
+            "ttl": 121,
+            "rtt": 255.168
           }
         ],
-        "pkt_rcv": 10,
-        "pkt_snt": 10,
-        "min_rtt": 56.477,
-        "max_rtt": 67.85,
-        "avg_rtt": 60.926,
-        "std_rtt": 3.717,
-        "error": null
-      },
-      {
-        "target": "10.41.0.1",
-        "sequence": [
-          {
-            "seq": 10,
-            "ttl": 64,
-            "rtt": 107.01
-          }
-        ],
-        "pkt_rcv": 10,
-        "pkt_snt": 10,
-        "min_rtt": 48.734,
-        "max_rtt": 107.01,
-        "avg_rtt": 58.518,
-        "std_rtt": 16.43,
-        "error": null
-      },
-      {
-        "target": "163.7.134.112",
-        "sequence": [
-          {
-            "seq": 10,
-            "ttl": 54,
-            "rtt": 334.934
-          }
-        ],
-        "pkt_rcv": 10,
-        "pkt_snt": 10,
-        "min_rtt": 327.153,
-        "max_rtt": 379.311,
-        "avg_rtt": 338.438,
-        "std_rtt": 14.347,
-        "error": null
+        "pkt_rcv": 1,
+        "pkt_snt": 1,
+        "min_rtt": 255.168,
+        "max_rtt": 255.168,
+        "avg_rtt": 255.168,
+        "std_rtt": 0,
+        "failure": null
       }
     ],
     "requests": [
-      {
-        "failure": null,
-        "request": {
-          "method": "GET",
-          "url": "https://api.ipify.org/?format=json",
-          "headers": {
-            "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
-            "accept-language": "en-US,en;q=0.9",
-            "user-agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"
-          }
-        },
-        "response": {
-          "code": 200,
-          "headers": {
-            "connection": "keep-alive",
-            "content-length": "23",
-            "content-type": "application/json",
-            "date": "Fri, 21 Oct 2022 09:12:18 GMT",
-            "server": "Cowboy",
-            "vary": "Origin",
-            "via": "1.1 vegur"
-          },
-          "body": {
-            "data": "eyJpcCI6IjIxMi44My4xODIuMTI3In0=",
-            "format": "base64"
-          },
-          "body_is_truncated": false,
-          "x_body_length": 23,
-          "x_body_is_utf8": true
-        },
-        "t": 0.558008438,
-        "started": 8.3589e-05,
-        "oddity": ""
-      }
-    ]
-  }
-}
+    // ...
+    ],
+    "minivpn_version": "(devel)",
+    "obfs4_version": "(devel)",
+    "success_handshake": true,
+    "success_icmp": true,
+    "success_urlgrab": true,
+    "success": true
+},
 ```
 
 where:
@@ -253,7 +211,8 @@ where:
   should match). However, the authoritative source is the entry in the
   test-keys.
 - `provider` (`string`) is the entity that manages the endpoints. If it's not a
-  provider known to OONI, it should be marked as "unknown".
+  provider known to OONI, it should be marked with a prefix starting with
+  "unknown".
 - `vpn_protocol` (`string`): should be `openvpn`.
 - `transport` (`string`): underlying transport used by OpenVPN, one of `tcp|udp`,
 - `remote` (`string`): IP address and port for the remote endpoint (`ipaddr:port`).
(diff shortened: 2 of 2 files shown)


nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    "provider": "riseup",
+    "vpn_protocol": "openvpn",
+    "transport": "tcp",
+    "remote": "51.158.144.32:80",
+    "obfuscation": "none",
+    "bootstrap_time": 0.37221659,
+    "minivpn_version": "(devel)",
+    "obfs4_version": "",
+    "success": true
+    "stages": [
```
hellais, line 126: Would it perhaps make sense to use the `network_events` data format (https://github.com/ooni/spec/blob/master/data-formats/df-008-netevents.md) for these sorts of events?

It seems like it fits pretty well within that model. You could use `transaction_id` in place of `op_id`.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    "tcp_connect": {
+      "ip": "51.158.144.32",
+      "port": 80,
+      "status": {
+        "failure": null,
+        "success": true
+      },
+      "t0": 0.000239501,
+      "t": 0.045310671,
+      "transaction_id": 1
```
hellais, line 147: What is the relationship between this `transaction_id` and other parts of the document?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    },
+    "failure": null,
+    "error": null,
+    "pings": [
```
hellais, line 151: I'm not sure it's a problem, but we might want to think about the fact that we currently have 3 different structures for the key `pings` (see: https://github.com/ooni/spec/blob/master/nettests/ts-033-tlsping.md & https://github.com/ooni/spec/blob/master/nettests/ts-032-tcpping.md & https://github.com/ooni/spec/blob/master/nettests/ts-031-quicping.md).

For example we might care to disambiguate all these different pings by using a clearly distinct key (ex. `icmp_pings`) so that somebody parsing the data isn't lead into thinking they are the same thing.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+          "url": "https://api.ipify.org/?format=json",
+          "headers": {
+            "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+            "accept-language": "en-US,en;q=0.9",
+            "user-agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"
+          }
+        },
+        "response": {
+          "code": 200,
+          "headers": {
```
hellais, line 223: Is `headers_list` ommitted here for brevity, or does it in fact not generate `headers_list`, but only the `headers` map?

I checked the output sample and could not find `headers_list` there either.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+          "body": {
+            "data": "eyJpcCI6IjIxMi44My4xODIuMTI3In0=",
+            "format": "base64"
+          },
+          "body_is_truncated": false,
+          "x_body_length": 23,
+          "x_body_is_utf8": true
+        },
+        "t": 0.558008438,
+        "started": 8.3589e-05,
```
hellais, line 241: Shouldn't this be `t0`?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+- `remote` (`string`): IP address and port for the remote endpoint (`ip:port`).
+- obfuscation": "none",
+- `bootstrap_time` (`float`) is the time, in seconds, to bootstrap a VPN connection,
+- `success` (`bool`) whether all the stages in the experiment were successful,
+- `minivpn_version` (`string`) contains the version of the `minivpn` library
+  that is used in the ooni probe build used for the experiment,
+- `obfs4_version` (`string`) contains the version of the `obfs4` library
+  that is used in the ooni probe build used for the experiment,
+- `failure` (`string`; nullable) conforms to `df-007-errors`,
+- `error` (`string`; nullable) raw error message, conforms to `df-007-errors`,
```
hellais, line 268: What's the difference between `failure` and `error`?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+- `tcp_connect`: info about the TCP handshake (only when `transport` == `tcp`). Conforms to `df-005-tcpconnect`.
+- `pings` is an array containing the result for a series of `icmp` pings through the tunnel:
+    - `target` (`string`): the IP the ICMP trains were targeting,
+    - `sequence` is an aray with the ping responses:
+        - `seq` (`int`): the sequence number for this response packet,
+        - `rtts`: (`float`): raw rtt value,
+        - `ttls`: (`int`): raw ttl value,
+    - `pkt_rcv` (`int`): how many packets were received,
+    - `pkt_snt` (`int`): how many packets were sent,
+    - `min_rtt` (`float`): the min value for all the pings,
```
hellais, line 282:
```suggestion
    - `min_rtt` (`float`): the minimum rtt value for all the pings,
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    - `max_rtt` (`float`): the max value for all the pings,
```
hellais, line 283:
```suggestion
    - `max_rtt` (`float`): the maximum rtt value for all the pings,
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    - `avg_rtt` (`float`): the avg value for all the pings,
```
hellais, line 284:
```suggestion
    - `avg_rtt` (`float`): the average of the rtt for all the pings towards this target,
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    - `std_rtt` (`float`): the std value for all the pings,
```
hellais, line 285:
```suggestion
    - `std_rtt` (`float`): the standard deviation of the rtt for all the pings towards this target,
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,710 @@
...
+    - `error` (`string`; nullable): any error during the ping operation. Conforms to `df-007-errors`.
```
hellais, line 286: For consistency with other data formats, this should maybe be called `failure` (unless it's something different, see: https://github.com/ooni/spec/pull/255/files#r1018916989)?

nettests/ts-039-openvpn.md
ainghazal: It makes sense - @bassosimone had previously recommended to mimic more closely that model. as per the df-008 operations are more restricted, but I guess I can specify that. Looking at, for instance, ts-029-signal I think it also makes sense to include a `failed_operation` field (int) so that we can easily track what's the failed state.

ainghazal: will change the key `stages` to `network_events` too.

nettests/ts-039-openvpn.md
ainghazal: I think this is always to be 1, since it's getting the first TCPConnect (*model.ArchivalTCPConnectResult):

`trace.FirstTCPConnectOrNil()`

https://github.com/ooni/probe-cli/blob/8dd1ae28a3593fcb7d083c0f3d9fb1263e648159/internal/measurexlite/dialer.go#L133

nettests/ts-039-openvpn.md
ainghazal: makes sense to rename as `icmp_pings`. it's possible that we end up internalizing the helper library too (it lives under github.com/ooni/minivpn/extras right now, but it needs some refactor to accomodate the differences between openvpn and wg tunnels).

nettests/ts-039-openvpn.md
ainghazal: see comment above; whatever logic is in use is coming directly from the archivalHTTP_ implementation. For what I see, t0 is basically `requests[0].started`

nettests/ts-039-openvpn.md
ainghazal: fixed

nettests/ts-039-openvpn.md
ainghazal: fixed

nettests/ts-039-openvpn.md
ainghazal: fixed

nettests/ts-039-openvpn.md
ainghazal: I think I was initially following the vanilla tor spec, but now I see error was there for backward compatibility reasons. I will remove in favor of failure.

ainghazal:
Pushed a new commit addressing comments. I still have to ask @bassosimone for possible replacement of the several measurement archivals in use (I might be using outdated operation wrappers).

nettests/ts-039-openvpn.md
hellais: We discussed that it might be that the older measurement library is being used here and that's why some fields are different/missing.

TODO: check with @bassosimone to know if that is the case.

ainghazal: @bassosimone confirmed I should be using `internal/model.ArchivalXXX` structs instead, since `measurex` is being deprecated. I'll update the PR.

ainghazal:
bumped to version 0.0.16; added ability to fetch arbitrary URLs.

data-formats/df-001-httpt.md
```diff
@@ -122,6 +122,10 @@ are multiple values for the same header key. See below the definition of
 - `tor` (`TorInfo`): this is an object containing information on the
 instance of `tor` that we may be using for measuring.
 
+- `vpn` (`VPNInfo`): similarly to `tor`, this is an object containing
```
bassosimone, line 125:
```suggestion
- `vpn` (`VPNInfo`; optional): similarly to `tor`, this is an object containing
```

We dont' want to include this field for every measurement like we (sadly) do for `tor`. We can also probably make `tor` optional without creating too much damage, but that would be out of scope here.

data-formats/df-001-httpt.md
```diff
@@ -222,6 +226,25 @@ if we're not using tor or this information is not available.
...
+```
+
+- `exit_ip` (`string`; nullable): the egress IP for the VPN gateway, or `null`
+  if we're not using a VPN or this information is not available.
+
+- `provider` (`string`; nullable): the name of the provider to which this VPN
+  gateway belongs to, or `null` if we're not using a VPN or this information is
+  not available.
+
+- `is_vpn` (`bool`): true if we're using a VPN, false otherwise.
```
bassosimone, line 247: If we're making the whole structure optional, we can avoid making its internal fields nullable. Likewise, we can avoid having a field called `is_vpn` that would always be `true`. (Historically, the tor structure is a such because the precursor of web connectivity was sending the same request over tor and cleartext, hence the data format ended up having a `tor` structure that we carry along in any measurement.)

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
+# Specification version number
+
+2022-10-20
```
bassosimone, line 3: We need to update this to the date of $today

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+
+Altogether, they constitute a basic subset of the valid configuration
+parameters for the reference openvpn implementation (`man 8 openvpn`).
+
+The input `URI` encodes information for the protocol (`openvpn`, in this case),
+provider, transport, obfuscation and the remote address.
+
+```
+vpn://{provider_name}.{protocol}/?transport=tcp&obfuscation=obfs4&address={ip}:{port}
+```
```
bassosimone, line 33: Please add an unordered list here where you explain the semantics of each parameter, like we normally do when you describe data structures.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+options to the experiment, encoded as a base64 string. We use the `Safe` prefix
+to signal to the engine that we don't want those options to be exported and
+indexed by the OONI backend:
+
+```
+"SafeCa": "base64:deadbeef",
+"SafeCert": "base64:deadbeef",
+"SafeKey": "base64:deadbeef",
+```
+
```
bassosimone, line 45:
```suggestion
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+The ability to detect if a working OpenVPN connection can be established with a given host.
+
+# Expected inputs
+
+There are two categories of input to this experiment. Some of them are given as
+part of the input `URI` for the experiment, and some others are passed as
+options.
+
+Altogether, they constitute a basic subset of the valid configuration
+parameters for the reference openvpn implementation (`man 8 openvpn`).
```
bassosimone, line 26: I think this section would be more readable if you would create subsections for URL and other options.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+
+The credentials to authenticate against such endpoint need to be passed as
+options to the experiment, encoded as a base64 string. We use the `Safe` prefix
+to signal to the engine that we don't want those options to be exported and
+indexed by the OONI backend:
+
+```
+"SafeCa": "base64:deadbeef",
+"SafeCert": "base64:deadbeef",
+"SafeKey": "base64:deadbeef",
```
bassosimone, line 43: I think it would be beneficial to explicit where these options would be set. Perhaps this could be helped by showing a subset of the toplevel structure including only the options and by referring to the base spec.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+```
+
+
+Other valid options, that will be passed to the openvpn invocation, are:
+
+- `cipher` (one of: `AES-256-GCM`, `AES-128-GCM`, `AES-256-CBC`, `AES-128-CBC`).
+- `auth` (one of: `SHA1`, `SHA512`).
```
bassosimone, line 50: I think you should explain with one sentence each of the possible options. Also beware that, IIUC, we need to use camel case names for options here.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+
+Additionally, the user can specify any amount of comma-separated urls that
+wants to be fetched through the tunnel. 
+
+- `URLs` (`string`, optional).
```
bassosimone, line 55: This is where I am sad that the data format only allows a list of strings. However, I the minioonirun v2 spec we allow the options to be a map from string to any. I think something is wrong in the base data format used for options and we should instead allow a map from string to any.

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+
+
+# Test description
+
+There are, at the moment, three distinct stages in the OpenVPN experiments:
+`tunnel initialization`, `icmp-ping`, and `urlgrabber`.
```
bassosimone, line 61: Is there a reason why you're using a space separator and a dash separator inconsistently?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+
+## 1. Tunnel initialization
+
+An OpenVPN connection tries, sequentially, the following steps:
+
+1. Initializes cryptographic material (local only, not expected to fail).
```
bassosimone, line 67: Can you add an empty line between each point thus facilitating vi-like `{` and `}` navigation?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+2. Creates an UDP or TCP socket and initializes the control channel (no network
+   communication in the case of UDP)
+3. Sends a client `hard-reset` packet, and waits for a `hard-reset` reset
+   packet from the server.
+4. Performs the TLS handshake over the control channel (can return a handshake
+   error, including a timeout if a reply is not received after a reasonable
+   amount of time).
+5. Data channel initialization: client sends a control packet (expects a server
+  ACK), and a request for the server to push options. Upon receiving options
+  from the server, the data channel is initialized and ready to use.
```
bassosimone, line 77:
```suggestion
   ACK), and a request for the server to push options. Upon receiving options
   from the server, the data channel is initialized and ready to use.
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+Request packets to different targets, and wait for their responses.
+
+We gather statistics about TTL, RTT (measured in milliseconds), and packet loss.
+
+By default, this experiment uses `n=10`.
+
+As reference targets, we measure one cloud provider, the tunnel gateway itself,
+and a box in a known geolocation.
+
+## 3. urlgrabber
```
bassosimone, line 96:
```suggestion
## 3. URLGrabber
```

Suggesting because above you have ICMP-Pings

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+geolocation service via an http `GET` that retrieves a json payload. We also
+fetch a truncated payload from a well-known website that should be enough to
+capture the language code in the html document.
+
+# Expected output
+
+## Parent data format
+
+We include data from the following parent formats:
+
```
bassosimone, line 108:
```suggestion
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+{
+    "provider": "riseup",
+    "vpn_protocol": "openvpn",
+    "transport": "tcp",
+    "remote": "198.252.153.109:443",
+    "obfuscation": "obfs4",
+    "bootstrap_time": 1.5671859910000001,
+    "network_events": [
+      {
+        "transaction_id": 0,
```
bassosimone, line 130: You are not documenting that you are using the network-events data format. Please modify the code such that we are not using the `transaction_id` for a different purpose than the one for which it is usually used.

I recommend you introduce your own `x_something` code for this experiment to keep the number that tells you the state of the handshake in which you are (or just drop the number?).

I think we should also clarify what happens in case of error. Does the failure in, say, stage 3 causes a `failure` to be written in the corresponding `network_events` stage?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+        "pkt_snt": 1,
+        "min_rtt": 255.168,
+        "max_rtt": 255.168,
+        "avg_rtt": 255.168,
+        "std_rtt": 0,
+        "failure": null
+      }
+    ],
+    "requests": [
+    // ...
```
bassosimone, line 202: Set the data type in the ``` using JavaScript. This will make comments render without errors.

bassosimone: Maybe you can show a single request for clarity?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+    ],
+    "minivpn_version": "(devel)",
```
bassosimone, line 204: I am not super sold about emitting a non version number as the version

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+    "success": true
+},
+```
+
+where:
+
+- `provider`, `vpn_protocol`, `transport`, `remote` and `obfuscation` are
+  parameters extracted by the probe from the input `URI` (and as such, they
+  should match). However, the authoritative source is the entry in the
+  test-keys.
```
bassosimone, line 218: Please add an empty line between each entry

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+  OpenVPN handshake, conforming to `df-008-netevents`:
+    - `transaction_id` (`int`): sequential integer for the operation.
+    - `operation` (`string`): the name for the operation.
+    - `t` (`float`): time, in seconds, for the event marking this operation.
+- `last_handshake_transaction_id` (`uint8`) integer that corresponds to the
+  transaction id of the last received OpenVPN handshake event.
+- `tcp_connect`: info about the TCP handshake (only when `transport` == `tcp`). Conforms to `df-005-tcpconnect`.
+- `icmp_pings` is an array containing the result for a series of `icmp` pings through the tunnel:
+    - `target` (`string`): the IP the ICMP trains were targeting,
+    - `sequence` is an aray with the ping responses:
```
bassosimone, line 249:
```suggestion
    - `sequence` is an array with the ping responses:
```

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+- `provider` (`string`) is the entity that manages the endpoints. If it's not a
+  provider known to OONI, it should be marked with a prefix starting with
+  "unknown".
+- `vpn_protocol` (`string`): should be `openvpn`.
+- `transport` (`string`): underlying transport used by OpenVPN, one of `tcp|udp`,
+- `remote` (`string`): IP address and port for the remote endpoint (`ipaddr:port`).
+- `obfuscation` (`string`): type of obfuscation in use. Currently known
+  obfuscation schemes include `none` and `obfs4`,
+- `bootstrap_time` (`float`) is the time, in seconds, to bootstrap a VPN connection,
+- `success` (`bool`) whether all the stages in the experiment were successful,
```
bassosimone, line 228: can you consistently use `.` instead of mixing `.` and `,` to signal the end of an item?

nettests/ts-039-openvpn.md
```diff
@@ -0,0 +1,688 @@
...
+comparison to the reference OpenVPN implementation. While work is being done to
+close the gap between the two implementations, a motivated adversary can in
+theory infer the usage of probing activity by means of implementation quirks.
+However, probing activity is already evident because of the traffic patterns
+and timing, so this should not be a high-impact concern in practice.
+
+# Packet capture considerations
+
+This test does not capture packets by default.
+
```
bassosimone, line 667:
```suggestion
```

bassosimone (commented):
Thanks for writing this spec! 🙌 

I have provided comments and suggestions!

data-formats/df-001-httpt.md
ainghazal: added

nettests/ts-039-openvpn.md
ainghazal: done

nettests/ts-039-openvpn.md
ainghazal: done

nettests/ts-039-openvpn.md
ainghazal: yea it seemed weird. I blindly followed cargo-cult while doing this. a new revision of the base format should be possible, isn't it?

nettests/ts-039-openvpn.md
ainghazal: probably some weird bias towards what a sintagm and a modificator are, but you're right. removing the dash.

nettests/ts-039-openvpn.md
ainghazal: https://github.com/ooni/spec/pull/255#discussion_r1018902876 I will revert to use `x_operation_id`. right now there is no failure inside the `network_event` (the failure is stored as the experiment `failure`, but it looks like this is the right thing to do).

nettests/ts-039-openvpn.md
ainghazal: this is annoying yes. it's a side-effect of using `go.work` locally for overrides. I'll investigate if I can fetch the original version somehow.

nettests/ts-039-openvpn.md
ainghazal: there's a request below, in "example output sample". do you think it's still needed here?

ainghazal:
closing now that we merged #293

PR #278, opened 2023-10-10
Update OONIRun spec

FedericoCeratto:
Updated based on the current implementation.
https://github.com/ooni/backend/pull/678
https://github.com/ooni/backend/pull/699
https://github.com/ooni/backend/pull/702

### DIFF
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -329,7 +329,7 @@ dashboard as follows.
 
 In order to support the above workflow the OONI API needs to support the following operations:
 
-* CREATE a new OONI Run link, returning the OONI Run link ID (see 4.1)
+* CREATE a new OONI Run link or a translation or other changes (see 4.1)
 
 * ARCHIVE an existing OONI Run link (see 4.2)
 
@@ -346,21 +346,31 @@ A certain OONI Run link can rendered ineffective by setting the `is_archived`
 flag to true. In this case the OONI Run link still remains available, but it
 will not lead to tests being initiated.
 
+Various API responses contain a response format version. It is an incremental
+integer under the key "v". It is used to identify changes in the response format
+and enable non-breaking rollouts of updates of the API and the web UIs.
+
 ## 4.1 CREATE a new OONI Run link
 
-This creates a new oonirun link or a new version for an existing one.
+This creates a new oonirun link or a new version or translation for an existing one.
 
-Then an OONI Run link is first created a new ID is assigned. In order to create
-new versions for an existing OONI Run link the argument `id` must be passed.
+Then an OONI Run link is first created a new ID is assigned by the API. In order to create
+new versions for an existing OONI Run link the argument `ooni_run_link_id` must be passed.
 Creating a new version may involve adding or changing translations, adding or
 removing tests, editing targets of existing ones or making changes to the OONI Run
 link metadata.
 
+The API allows keeping history of previous versions in case users want to fetch an older
+version instead of the latest one.
+
+The fields `descriptor_creation_time` and `translation_creation_time` are automatically
+updated by the API as needed when a new version or a new translation are created.
+
 The web UI should discourage users from making changes to the title, icon and
-descriptions of OONI Run links as to not confused volunteers that have installed
+descriptions of OONI Run links as to not confuse volunteers that have installed
 a link.
 
-This operation will be performed by a logged in user that is interested in
+This operation will be performed by a logged-in user that is interested in
 performing an OONI Run link based measurement campaign.
 
 It is outside of the scope of this document to specify how registration and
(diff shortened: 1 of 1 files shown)


backends/bk-005-ooni-run-v2.md
```diff
@@ -395,109 +415,97 @@ following JSON body:
 
 ```JavaScript
 {
-"ooni_run_link_id": "", // `string` OONI Run link identifier.
-
-"title": "",
 
-"description": "",
+"oonirun_id": "", // `string` OONI Run link identifier.
```
hellais, line 419: For consistency can we do a `s/oonirun_id/ooni_run_link_id` in the backend and frontend?

backends/bk-005-ooni-run-v2.md
```diff
@@ -395,109 +415,97 @@ following JSON body:
...
 
-"author": "",
-
-// [... rest of the OONI Run link payload]
+"v": 1,
 
 }
 ```
 
-## 4.2 UPDATE an existing OONI Run link
```
hellais, line 411: What happened to updates for OONI Run links? I believe it is supported, so why was the update text removed?

backends/bk-005-ooni-run-v2.md
```diff
@@ -506,16 +514,17 @@ As such, this request does not require any authentication.
...
 To retrieve an OONI Run link descriptor, the client issues a request compliant with:
 
-`GET /api/v1/ooni_run/{ooni_run_link_id}`
-
-### Response status code
+`GET /api/_/ooni_run/list` with optional arguments:
+ - "only_latest": <boolean> // Return only the latest version of each OONI Run link
+ - "include_archived": <boolean> // Include archived links
+ - "only_mine": <boolean> // Show only links owned by the current user
+ - "ids": <string> // Comma-separated OONI Run link ID numbers to filter for
```
hellais, line 521: How are these "arguments" passed? Does this mean these are URL **parameters**? If that is the case, then we should clarify that's what we mean specify how they are passed.

hellais (changes requested):
Some changes need to be done for clarity. The UPDATE method sections is missing. We should, for consistency, change the name of `oonirun_id` to `ooni_run_link_id` so it's consistent with what we use elsewhere.

backends/bk-005-ooni-run-v2.md
```diff
@@ -395,109 +415,97 @@ following JSON body:
...
-"description": "",
+"descriptor_creation_time": "<string>" // Descriptor creation time
 
-"author": "",
+"mine": "<boolean" // True if the descriptor belongs to the logged-in user.
 
-// [... rest of the OONI Run link payload]
+"translation_creation_time": "<string>" // Translation creation time
 
+"v": 1 // Response format version
```
hellais, line 502: Since this appears in many parts of the spec, we probably should make a table at the beginning of the doc explaining this `v` option and include a list of current version (which now starts from 1)

backends/bk-005-ooni-run-v2.md
hellais: Why are we passing the `ooni_run_link_id` as part of the create operation?

backends/bk-005-ooni-run-v2.md
FedericoCeratto: Added.

backends/bk-005-ooni-run-v2.md
FedericoCeratto: When support for translations + URL updates have been added to the OONI Run links we implemented them as appending new records to the database table in order:
- to keep history of previous versions in case users want to fetch an older version instead of the latest one
- to allow incremental improvements of translations
This is implemented in https://github.com/ooni/backend/pull/699 - also see https://docs.google.com/document/d/1tsik0DZLOgYHh01RsLWWXp3ZHhHB_C07j8dcaMbGB9M/edit
On the database side rather than updating an existing OONI Run we are creating either a new version or a new translations, hence every action can be done with the `/api/_/ooni_run/create` path.
To add more items related to an already existing OONI Run link the Web UI passes `ooni_run_link_id` .
When absent the API generates a new `ooni_run_link_id` value.

backends/bk-005-ooni-run-v2.md
FedericoCeratto: See previous reply.

backends/bk-005-ooni-run-v2.md
```diff
@@ -344,14 +358,20 @@ authentication should be handled.
 When you `CREATE` a new OONI RUN link, the client sends a HTTP `POST`
 request conforming to the following:
 
-`POST /api/v1/ooni_run`
+`POST /api/_/ooni_run/create` with optional `id` argument
```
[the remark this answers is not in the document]
FedericoCeratto: See previous reply.

backends/bk-005-ooni-run-v2.md
FedericoCeratto: They are described in https://api.ooni.io/apidocs/#/default/get_api___ooni_run_list - I'll update the PR

hellais:
This is superseeded by https://github.com/ooni/spec/pull/292

PR #292, opened 2024-02-23
Update OONI Run v2 spec following refactoring: https://github.com/oon…

hellais:
…i/backend/pull/805

This PR superseeds: https://github.com/ooni/spec/pull/278

### DIFF
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -1,8 +1,8 @@
 # OONI Run v2 specification
 
-* author: Norbel Ambanumben, Arturo Filastò
-* version: 2023.10.18
-* status: _draft_
+-   author: Norbel Ambanumben, Arturo Filastò
+-   version: 2024.02.23
+-   status: release-candidate
 
 This document provides a functional specification for OONI Run.
 
@@ -357,17 +357,22 @@ authentication should be handled.
 When you `CREATE` a new OONI RUN link, the client sends a HTTP `POST`
 request conforming to the following:
 
-`POST /api/v1/ooni_run`
+`POST /api/v2/oonirun`
 
 ```JavaScript
 {
 "name": "", // (required) `string` is the display name for the OONI Run link
 
-"description": "", // (optional) `string` describing the scope of this OONI Run link system
+"description": "", // (required) `string` describing the scope of this OONI Run link system
+
+"author": "", // (required) `string` name of the creator of this OONI Run link
+
+"name_intl": {"it": ""}, // (optional) `string` is the display name for the OONI Run link
+
+"description_intl": {"it": ""}, // (optional) `string` describing the scope of this OONI Run link system
 
 "icon": "", // (optional) `string` the ID of any icon part of the OONI icon set
 
-"author": "", // (optional) `string` name of the creator of this OONI Run link
 
 "nettests": // `array` provides a JSON array of tests to be run.
    [
@@ -408,7 +413,6 @@ following JSON body:
 
 ```JavaScript
 {
-"ooni_run_link_id": "", // `string` OONI Run link identifier.
 
 "title": "",
 
@@ -417,6 +421,17 @@ following JSON body:
 "author": "",
 
 // [... rest of the OONI Run link payload]
+
+// Additional fields that are added by the backend are:
+
+"oonirun_link_id": "", // `string` OONI Run link identifier.
+"is_archived": false,
+"date_created": "",
+"date_updated": "",
+"creator_account_id": "",
+"revision": 1,
+"is_mine": false,
+"v": 1,
 
 }
 ```
@@ -439,37 +454,19 @@ a link.
 
 ### Request
 
-To update an OONI Run Link, the client issues a request compliant with:
+To update an OONI Run Link, the client issues a request compliant the same as the create request.
 
-`PUT /api/v1/ooni_run/{ooni_run_link_id}`
+Below we list the extra fields that are settable from the edit request that are
+not settable during CREATE.
+
+`PUT /api/v2/oonirun/{ooni_run_link_id}`
 
 ```JavaScript
 {
-"name": "", // (required) `string` is the display name for the OONI Run link
+   // See create for full semantics
 
-"description": "", // (optional) `string` describing the scope of this OONI Run link system
-
-"icon": "", // (optional) `string` the ID of any icon part of the OONI icon set
-
-"author": "", // (optional) `string` name of the creator of this OONI Run link
-
-"nettests": // `array` provides a JSON array of tests to be run.
-   [
-      {
-         "inputs": [
-            "https://example.com/",
-            "https://ooni.org/",
-            "https://torproject.org/"
-         ],
-         "options": {
-            "HTTP3Enabled": true,
-         },
-         "test_name": "web_connectivity"
-      },
-      {
-         "test_name": "dnscheck"
-      }
-   ]
+   // In addition to those fields, it's possible to specify the is_archived to set a link as archived.
+   "is_archived": false,
 }
 ```
 
@@ -477,7 +474,7 @@ To update an OONI Run Link, the client issues a request compliant with:
 
 Upon receiving this request, the OONI Run backend:
 
-1. SHOULD check whether the `${ooni_run_id}` exists and they have permission to
+1. SHOULD check whether the `${oonirun_link_id}` exists and they have permission to
    edit it and reject the request with a `4xx` status otherwise.
 
 2. SHOULD reject the request with a `4xx` if the JSON does not
@@ -495,7 +492,7 @@ following JSON body:
 
 ```JavaScript
 {
-"ooni_run_link_id": "", // `string` OONI Run link identifier.
+"oonirun_link_id": "", // `string` OONI Run link identifier.
 
 "title": "",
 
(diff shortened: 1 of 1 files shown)


backends/bk-005-ooni-run-v2.md
```diff
@@ -357,17 +357,22 @@ authentication should be handled.
...
-`POST /api/v1/ooni_run`
+`POST /api/v2/oonirun`
 
 ```JavaScript
 {
 "name": "", // (required) `string` is the display name for the OONI Run link
 
-"description": "", // (optional) `string` describing the scope of this OONI Run link system
+"description": "", // (required) `string` describing the scope of this OONI Run link system
+
```
majakomel, line 367: If we want to include all fields here, those are missing: `short_description` (optional), `short_description_intl` (optional).
We also added `color` with hex value, to use as a card background in the app.
The last one is `expiration_date` as YYYY-MM-DD

backends/bk-005-ooni-run-v2.md
```diff
@@ -439,45 +454,27 @@ a link.
...
-            "HTTP3Enabled": true,
-         },
-         "test_name": "web_connectivity"
-      },
-      {
-         "test_name": "dnscheck"
-      }
-   ]
+   // In addition to those fields, it's possible to specify the is_archived to set a link as archived.
+   "is_archived": false,
```
majakomel, line 469: I think I completely removed the button to manage the "archived" flag since we added `expiration_date` field and idea was to manage the expiration/archiving only via `expiration_date` in the interface. Do we want to allow updates to `is_archived` or should only the backend make changes to it?

backends/bk-005-ooni-run-v2.md
```diff
@@ -357,17 +357,22 @@ authentication should be handled.
...
 
-`POST /api/v1/ooni_run`
+`POST /api/v2/oonirun`
 
 ```JavaScript
 {
 "name": "", // (required) `string` is the display name for the OONI Run link
 
-"description": "", // (optional) `string` describing the scope of this OONI Run link system
+"description": "", // (required) `string` describing the scope of this OONI Run link system
```
majakomel, line 366: Should we also include the max length for `name`, `short_description`, `description`? https://github.com/ooni/ooni.org/issues/1547, there's also TODO on line 106 to recommend a maximum length.

backends/bk-005-ooni-run-v2.md
```diff
@@ -357,17 +357,22 @@ authentication should be handled.
...
+
+"author": "", // (required) `string` name of the creator of this OONI Run link
```
majakomel, line 368: Author can now also be empty, if user doesn't want to include their email. 

```suggestion
"author": "", // `string` email address of the creator of this OONI Run link
```

backends/bk-005-ooni-run-v2.md
hellais: Good point. I have indeed added minimum length validation, to the backend, but not max length. Maybe we should add that too.

Do we have an idea of what the values for max_lenght should be?

backends/bk-005-ooni-run-v2.md
hellais: Good point. I'm actually missing support for `color` and `expiration_date` since I didn't see it in the other PR for the spec changes, but I guess that's because the backend previously just treats the descriptor as opaque, while I am now doing more strict validation of it and defining the semantics upfront.

backends/bk-005-ooni-run-v2.md
```diff
@@ -357,17 +357,22 @@ authentication should be handled.
...
+
+"name_intl": {"it": ""}, // (optional) `string` is the display name for the OONI Run link
+
+"description_intl": {"it": ""}, // (optional) `string` describing the scope of this OONI Run link system
 
 "icon": "", // (optional) `string` the ID of any icon part of the OONI icon set
```
aanorbel, line 374:
```suggestion
"icon": "", // (optional) `string` the ID of any icon part of the OONI icon set

"color": "", // (optional) `string` the hex color to be used when displaying the test
```

backends/bk-005-ooni-run-v2.md
hellais: Let me think about it. You may be right that it's perhaps redundant since we have the expiration_date field, however it might be simpler in terms of logic to have it still be present as a boolean returned value so that the client doesn't need to have a reliable clock to validate if a descriptor is expired.

backends/bk-005-ooni-run-v2.md
```diff
@@ -418,6 +422,17 @@ following JSON body:
...
+// Additional fields that are added by the backend are:
+
+"oonirun_link_id": "", // `string` OONI Run link identifier.
+"is_archived": false,
+"date_created": "",
+"date_updated": "",
+"creator_account_id": "",
+"revision": 1,
+"is_mine": false,
+"v": 1,
```
aanorbel, line 434: Will this version be incremented whenever there is a change in the defined descriptor?

backends/bk-005-ooni-run-v2.md
```diff
@@ -418,6 +422,17 @@ following JSON body:
...
 // [... rest of the OONI Run link payload]
 
+// Additional fields that are added by the backend are:
+
+"oonirun_link_id": "", // `string` OONI Run link identifier.
+"is_archived": false,
+"date_created": "",
+"date_updated": "",
+"creator_account_id": "",
+"revision": 1,
```
aanorbel, line 432: I am not to clear on the difference between `revision` and `v`

backends/bk-005-ooni-run-v2.md
```diff
@@ -418,6 +422,17 @@ following JSON body:
 
 // [... rest of the OONI Run link payload]
 
+// Additional fields that are added by the backend are:
+
+"oonirun_link_id": "", // `string` OONI Run link identifier.
+"is_archived": false,
+"date_created": "",
+"date_updated": "",
+"creator_account_id": "",
```
aanorbel, line 431: I am not too clear on the usage `creator_account_id` and `author`

backends/bk-005-ooni-run-v2.md
majakomel: For GET requests I agree, I'm just not sure about users being able to update `is_archived`

backends/bk-005-ooni-run-v2.md
hellais: Honestly I think the version in the descriptor is a bit redudant. I just added it because it was used elsewhere, but I think that if we make breaking changes to the API we should bump the version in the path and not rely on clients expecting to read this.

The descriptors might change in way that are not breaking and I think it would be misleading do have to bump that always.

Would it be OK to just remove this?

backends/bk-005-ooni-run-v2.md
hellais: It explained in the section `GET /api/v2/oonirun/{oonirun_link_id}`, but I should add something here as well.

Basically `revision` is used to differentiate between different revisions of the same OONI Run link. Previously this was done by using the `updated_timestamp`, but now I think we ought to instead of use the `revision` as an ever increasing int value, which seems more robust and easy to use.

backends/bk-005-ooni-run-v2.md
aanorbel: We have to track this value somewhere. If we are using the path, then we can remove it from here.

backends/bk-005-ooni-run-v2.md
majakomel: I think it was never even exposed in the API, not using it anywhere - can be dropped

backends/bk-005-ooni-run-v2.md
majakomel: Yes, it includes `#` and 6 chars, but we could allow shortened version too

hellais: I would say we should probably only allow the long form since that's easier to handle across platforms (where we don't have CSS we need to worry about converting the short form into long form if we don't have native support for it).

backends/bk-005-ooni-run-v2.md
hellais: Are you reading the value of `v` somewhere and using it as part of some logic to do something?

backends/bk-005-ooni-run-v2.md
hellais: It has been dropped.

backends/bk-005-ooni-run-v2.md
hellais: The `is_archived` cannot be set by the user in the last version of the spec. The field called `is_expired` will be populated automatically and is a read-only properly that's not settable by the user.

Apps that need to know the state of a description should ignore the expiration and trust the `is_expired` flag (the clock of the device might be out of sync so future/past comparisons should not happen on device).

backends/bk-005-ooni-run-v2.md
hellais: We now support both `color` and `expiration_date`

backends/bk-005-ooni-run-v2.md
hellais: I spoke to maja and we decided it's probably best to analyze this in the context of the mobile app. In any case the authoritative source for this should be the api docs and not the spec: https://dataapi.prod.ooni.io/docs#/

backends/bk-005-ooni-run-v2.md
hellais: Fixed this comment.

mScott224 (approved):

PR #293, opened 2024-03-07
feat: add openvpn experiment spec

ainghazal:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2688
- [x] related ooni/probe-cli pull request: https://github.com/ooni/probe-cli/pull/1585

## Description

Add initial spec for the `openvpn` experiment.
This proposal supersedes a previous version (https://github.com/ooni/spec/pull/255).
ICMP Pings and URLGrabber over the tunnel are not included, this experiment just does OpenVPN handshakes at the moment.

### DIFF
--- a/nettests/ts-040-openvpn.md
+++ b/nettests/ts-040-openvpn.md
@@ -1,6 +1,6 @@
 # Specification version number
 
-2024-04-05
+2024-06-03
 
 * _status_: experimental.
 
@@ -11,6 +11,8 @@
 # Test preconditions
 
 * An internet connection
+* Configuration and Inputs for the given targets; provided by any means like
+  the OONI API, `miniooni` cli options, `oonirunv2` descriptors or Richer Input.
 
 # Expected impact
 
@@ -136,9 +138,13 @@ Where:
 The `openvpn_handshake` entry contains information about the result of every
 openvpn connection attempt performed during the experiment.
 
-- `bootstrap_time` (`float`): the total time until successful handshake or failure, relative to the beginning of the handshake (`t - t0`). Do note that, for TCP, the effective time should include the time for the TCP connection.
+- `bootstrap_time` (`float`): the total time until successful handshake or
+  failure, relative to the beginning of the handshake (`t - t0`). Do note that,
+  for TCP, the effective time should include the time for the TCP connection.
 
 - `endpoint` (`string`): a URI representing the probed endpoint. This is a different encoding than the input URI format.
+
+- `failure` (`string|null`): any failure returned by the operation.
 
 - `ip` (`string`): the IP address of the endpoint.
 
@@ -152,12 +158,6 @@ openvpn connection attempt performed during the experiment.
 - `openvpn_options`: a map from `string` to `string`, containing the relevant
    subset of the openvpn options used in the connection, for comparison
    purposes.
-
-- `status` summarizes the result and possible failures of the connection:
-
-   - `failure` (`string`): any failure returned by the operation.
-
-   - `success` (`boolean`): if true, it marks a sucessful tunnel initialization.
 
 - `t0` (`float`): the beginning of the openvpn handshake, in seconds, relative to `measurement_start_time`. In TCP
   mode, this is right after a successful TCP three-way-handhsake.
@@ -174,6 +174,7 @@ Example:
   {
     "bootstrap_time": 0.177719632,
     "endpoint": "openvpn://51.15.187.53:1194/tcp",
+    "failure": null,
     "ip": "51.15.187.53",
     "port": 1194,
     "transport": "tcp",
@@ -181,10 +182,6 @@ Example:
     "openvpn_options": {
       "auth": "SHA512",
       "cipher": "AES-256-GCM"
-    },
-    "status": {
-      "failure": "",
-      "success": true
     },
     "t0": 0.033876298,
     "t": 0.21159593,
@@ -583,6 +580,7 @@ The `network_events` array follows `df-008-netevents` semantics, with some diffe
       {
         "bootstrap_time": 0.177719632,
         "endpoint": "openvpn://51.15.187.53:1194/tcp",
+        "failure": null,
         "ip": "51.15.187.53",
         "port": 1194,
         "transport": "tcp",
@@ -590,10 +588,6 @@ The `network_events` array follows `df-008-netevents` semantics, with some diffe
         "openvpn_options": {
           "auth": "SHA512",
           "cipher": "AES-256-GCM"
-        },
-        "status": {
-          "failure": "",
-          "success": true
         },
         "t0": 0.033876298,
         "t": 0.21159593,
--- a/nettests/ts-040-openvpn.md
+++ b/nettests/ts-040-openvpn.md
@@ -1,0 +1,975 @@
+# Specification version number
+
+2024-03-07
+
+* _status_: experimental.
+
+# Specification name
+
+`openvpn`
+
+# Test preconditions
+
+* An internet connection
+
+# Expected impact
+
+The ability to detect if a working OpenVPN connection can be established with a given host.
+
+# Expected inputs
+
+There are two categories of input to this experiment. Some of them are given as
+part of the input `URI` for the experiment, and some others are passed as
+options.
+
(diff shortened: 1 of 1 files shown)


ainghazal:
For reference: https://explorer.ooni.org/m/20240307214346.366779_MX_openvpn_7df887fc7cd3836b

nettests/ts-040-openvpn.md
```diff
@@ -0,0 +1,975 @@
...
+  "resolver_ip": "172.253.251.38",
+  "resolver_network_name": "Google LLC",
+  "software_name": "miniooni",
+  "software_version": "3.21.0-alpha",
+  "test_keys": {
+    "success_all": true,
+    "success_any": true,
+    "inputs": [
+      "openvpn://51.15.187.53:1194/udp/?provider=riseup",
+      "openvpn://51.15.187.53:1194/tcp/?provider=riseup"
```
hellais, line 343: Is it perhaps not possible to promote these `inputs` keys to a top level `input` key and write inside of it a single entry?

How about instead of having a list that contains in it both `tcp` and `udp`, you write in the `input` key the address and provider name and specify which protocols are enabled by setting an `option`?

For example this could look like:
```javascript
{
  "input": "openvpn://51.15.187.53:1194/?provider=riseup",
  "options": [
      "UDPEnabled=true",
      "TCPEnabled=true"
  ]
}
```

This way we can maintain the semantic of `input` specifies the target of the measurement and integrate easily with the existing tooling, such as OONI Explorer, allowing users to search for things, while at the same allowing to parse which capability has been tested via the experiment by optionally parsing the `options` field.

That's consistent, for example, with what the URLGetter test does in setting the `HTTP3Enabled=true` option (see: https://explorer.ooni.org/m/20240321113140.297285_VE_urlgetter_23152e2fc6114c42)

nettests/ts-040-openvpn.md
```diff
@@ -0,0 +1,975 @@
...
+        "provider": "riseup",
+        "openvpn_options": {
+          "auth": "SHA512",
+          "cipher": "AES-256-GCM"
+        },
+        "status": {
+          "failure": "",
+          "success": true
+        },
+        "handshake_start_time": "2024-03-07T21:52:09.351623009+01:00",
```
hellais, line 895: The timestamp here should be formatted as UTC timestamp and using `Z` to indicate it's in the UTC timezone.

For example:

```
2024-03-07T22:52:09.351623009Z
```

hellais: It's also actually maybe a bit redundant to have also a `handshake_start_time` given that you are already recording this information, I assume, as part of the `t` value which is a monotonic clock that expresses the offset to the overall `measurement_start_time`

nettests/ts-040-openvpn.md
```diff
@@ -0,0 +1,975 @@
...
+          "auth": "SHA512",
+          "cipher": "AES-256-GCM"
+        },
+        "status": {
+          "failure": "",
+          "success": true
+        },
+        "handshake_start_time": "2024-03-07T21:52:10.051371358+01:00",
+        "t0": 0.222041219,
+        "t": 0.938619976,
```
hellais, line 918: I don't know if this is a real measurement, but assuming it is, it looks to me like there might be some issue (or maybe some misunderstanding on my end) of the values of `t0`, `t` and `bootstrap_time`. I would expect `bootstrap_time` to be equal to `t - t0`, while that is not the case in this particular measurement entry.

nettests/ts-040-openvpn.md
ainghazal: I've been trying to implement the separation of endpoint and transport.

However, unless I'm mistaken, the current semantics for Options are parameters that come from command line (or oonirun descriptors), and are set globally for a single run, not something that can be parametrized for a list of inputs.

My current implementation follows more or less the `loadOrQueryBackend` implementation in `engine.InputLoader`: https://github.com/ooni/probe-cli/blob/master/internal/engine/inputloader.go#L302 and so it fetches a list of URL-like "inputs".

I can convert these lists of inputs into an intermediate endpoint representation, but I think there'll be some friction in trying to modify the options for each input (I might be wrong). 

If for indexing purposes it doesn't *need* to be an Option, I could treat it like a per-endpoint parameter, in a similar way that it's done for Tor Targets: https://github.com/ooni/probe-cli/blob/master/internal/model/ooapi.go#L127 and then treat it as an internal "parameter" specific to the openvpn (or wg) experiment.

Conceptually, it would make sense to treat the IP address as the primary index for the endpoint, and pass tuples of (port, transport) as different variations to try for a single input:

```
api_input: {
  input: "openvpn://riseup.corp?address=51.15.187.53",
  params: [
    {port: 1194, transport: tcp},
    {port: 1194, transport: udp},
    {port: 443, transport: tcp},
    {port: 443, transport: udp},
}
```

In this way, and for a single IP endpoint input, the experiment would attempt 4 handhshakes, per port-transport combination (and they'd be indexed by the transaction ID in the respective test key entries).

If this facilitates analysis, let's go for it. 

It might be also possible to augment the current semantics of `Options` in the client to build them on a per-measurement basis. A simple way into this could be to have the experiment re-write options for each measurement Run (according to the combinations that we actually see make sense from the API response).

 

@hellais: is there something particular for the explorer/analysis end of things on having a parameter indexed as an Option, or could an internal parameter field also do?

ainghazal: FWIW: 

> Is it perhaps not possible to promote these inputs keys to a top level input key and write inside of it a single entry?

I just noticed that this review was based on an outdated spec. What the current code does is this:

1. Query the new endpoint in /v2/ooniprobe/vpn-config/riseup
2. Session/InputLoader fetches two things: 
    - A per-provider map of configuration options (including credentials)
    - A list of known endpoints to test.
3. Engine will run a single measurement PER INPUT, so with the current state of things in the integration branch there's ONE MEASUREMENT per port and transport.

What we're discussing here is to aggregate different transports (and potentially, ports) into a single measurement.

ainghazal: On the other hand, I do see some advantage to sticking to a single measurement measuring a single "thing", i.e., a given port/transport combination. 

For one, it flattens the observation, and it might make way cheaper to identify cases where we get evidence that we've got a longitudinal failure that might be attributable to the provider, and not network interference.

nettests/ts-040-openvpn.md
ainghazal: you're right, this was wrong. changed the semantics to t-t0 in the reviewed version.

nettests/ts-040-openvpn.md
ainghazal: changed as per suggestion.

nettests/ts-040-openvpn.md
```diff
@@ -0,0 +1,651 @@
...
+- `network_events` is an array containing timing for different stages of the
+   OpenVPN handshake, conforming to `df-008-netevents`. As in the case above,
+  `transaction_id` contains an index that allows to reference
+   events with the handshakes in the `openvpn_handshake` array. Do note that
+   timing information is relative to the beginning of each handshake.
+
+### openvpn_handshake
+
+The `openvpn_handshake` entry contains information about the result of every
+openvpn connection attempt performed during the experiment.
```
bassosimone, line 139:
```suggestion
openvpn connection attempt performed during the experiment:
```

nettests/ts-040-openvpn.md
```diff
@@ -0,0 +1,651 @@
...
+    - START: we've done the three-way handshake.
+    - SENT_KEY: we have sent the local part of the key_source2 random material.
+    - GOT_KEY: we have got the remote part of key_source2.
+    - ACTIVE: the control channel was established.
+    - GENERATED_KEYS: the data channel keys have been generated.
+
+    Some values mark inconsistent internal state:
+
+    - ERROR: protocol error.
+	- UNDEF: the state machine is in an ambiguous state.
```
bassosimone, line 219:
```suggestion
    - UNDEF: the state machine is in an ambiguous state.
```

bassosimone (approved):
💥

mScott224 (approved):

PR #297, opened 2024-11-07
Update spec to include changes to the echcheck test

hellais:
## Checklist

- [x] reference issue for this pull request: 
- [x] related ooni/probe-cli pull request: https://github.com/ooni/probe-cli/pull/1658

## Description

Please, insert here a more detailed description.


data-formats/df-006-tlshandshake.md
```diff
@@ -73,6 +74,14 @@ to verify the server's X.509 certificate. Note that, when this field contains an
 address rather than a domain name, the corresponding value is not included in the TLS
 ClientHello as described by [RFC 6066, Section 3](https://datatracker.ietf.org/doc/html/rfc6066#section-3);
 
+- `outer_server_name`: (`string`; optional): server name used in the OuterClientHello when [TLS ECH](https://www.ietf.org/archive/id/draft-ietf-tls-esni-22.html) is being used. When this is set, the `server_name` field indicates the field used inside of the encrypted client hello.
```
Em4E, line 77: Perhaps "ClientHelloOuter" to match [terminology](https://www.ietf.org/archive/id/draft-ietf-tls-esni-22.html#section-5.2) in the draft.

data-formats/df-006-tlshandshake.md
```diff
@@ -73,6 +74,14 @@ to verify the server's X.509 certificate. Note that, when this field contains an
...
+
+- `echconfig`: (`string`; optional): echconfig as defined in [TLS ECH
```
Em4E, line 79: Suggest "ECHConfig as defined in..." to match capitalization in the draft. `echconfig` as a key is fine.

data-formats/df-006-tlshandshake.md
```diff
@@ -73,6 +74,14 @@ to verify the server's X.509 certificate. Note that, when this field contains an
...
+  Spec](https://www.ietf.org/archive/id/draft-ietf-tls-esni-22.html#name-encrypted-clienthello-confi)
+base64 encoded as it would be presented inside of an SVCB HTTPS SvcParam as per
+[RFC9460](https://www.rfc-editor.org/rfc/rfc9460.html). In the event that only
+[GREASEd ECH](https://www.ietf.org/archive/id/draft-ietf-tls-esni-22.html#name-grease-psk) is being used, it will contain the string litteral `GREASE`.
```
Em4E, line 83: I haven't seen the term "GREASEd ECH" used elsewhere, only [GREASE ECH](https://www.ietf.org/archive/id/draft-ietf-tls-esni-22.html#name-grease-ech).

Typo "litteral".

nettests/ts-039-echcheck.md
```diff
@@ -29,9 +29,14 @@ the `input`.
 # Test description
 
 Before performing the test, this experiment will resolve the given target URL
-and establish a TCP connection. It will then attempt two TLS handshakes - one
-with an ECH extension, and another control handshake without an ECH extension
-present.
+and establish a TCP connection. It will then attempt three TLS handshakes - one
+with an ECH extension and the `public_name` (see: [ECH spec](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-22#section-6.1-6)) of the `echconfig` for the domain
+in the OuterClientHello, secondly ECH with a different `public_name` than that
```
Em4E, line 34: ClientHelloOuter

nettests/ts-039-echcheck.md
```diff
@@ -29,9 +29,14 @@ the `input`.
...
+advertised in the `public_name` field, finally a control handshake without an
+ECH extension present.
+
+The SNI used inside of the OuterClientHello can be distinguished by looking at
+the value of `outer_client_hello`.
```
Em4E, line 39: Is `outer_client_hello` supposed to be `outer_server_name` here?

nettests/ts-039-echcheck.md
```diff
@@ -29,9 +29,14 @@ the `input`.
 # Test description
 
 Before performing the test, this experiment will resolve the given target URL
-and establish a TCP connection. It will then attempt two TLS handshakes - one
-with an ECH extension, and another control handshake without an ECH extension
-present.
+and establish a TCP connection. It will then attempt three TLS handshakes - one
```
Em4E, line 32: Consider specifying that the three handshakes are done in a random order. This is to mitigate the case when the handshakes are not independent, such as when there is "residual censorship". Suppose handshake A causes blocking that lasts for 120 seconds, and handshake B does not cause blocking. If you always send in the order A, B, it will always look as if both handshakes were blocked. But if you occasionally also send B, A, you'll have some evidence that B does not result in blocking.

nettests/ts-039-echcheck.md
```diff
@@ -29,9 +29,14 @@ the `input`.
...
+with an ECH extension and the `public_name` (see: [ECH spec](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-22#section-6.1-6)) of the `echconfig` for the domain
```
Em4E, line 33: I don't see where the `echconfig` comes from. Is it from DNS, as usual, or is it e.g. hardcoded? Is encrypted DNS required?

Are there measurement keys that would record and interference that occurs during the process of looking up the `echconfig`? (Maybe there doesn't need to be, as long as there's an error message or something in the TLS connection keys.)

nettests/ts-039-echcheck.md
```diff
@@ -29,9 +29,14 @@ the `input`.
...
+in the OuterClientHello, secondly ECH with a different `public_name` than that
+advertised in the `public_name` field, finally a control handshake without an
+ECH extension present.
+
+The SNI used inside of the OuterClientHello can be distinguished by looking at
```
Em4E, line 38: ClientHelloOuter

nettests/ts-039-echcheck.md
hellais: At the moment all the test inputs are hardcoded.

 In the future we might either do the discovery of ECHConfig with DoH (which is what browsers do), but as you point out has the issue of interference happening on the lookup phase, or we might serve these inputs from the OONI backend service.

In any case I will add a note to the spec to mention these are hardcoded.

nettests/ts-039-echcheck.md
hellais: Yeah this is a good idea. I will see to modify the implementation to support this.

DecFox (approved):
LGTM!

PR #300, opened 2025-01-21
Add DDR experiment spec

frcroth:
## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/ooni/spec/blob/master/CONTRIBUTING.md)
- [ ] reference issue for this pull request: https://github.com/ooni/probe/issues/2819
- [ ] related ooni/probe-cli pull request: https://github.com/ooni/probe-cli/pull/1675
- [ ] If I changed a spec, I also bumped its version number and/or date

## Description

Please, insert here a more detailed description.

### DIFF
--- a/data-formats/df-002-dnst.md
+++ b/data-formats/df-002-dnst.md
@@ -212,7 +212,7 @@ that should be elapsed before the zone should be retried in case of failure.
 - `serial_number` (`int`; only for SOA answers): version number
 of the original copy of the zone.
 
-- `svcb`: (only for SCVB answers): the SVCB record, see below.
+- `svcb`: (`map[string]string`; only for SCVB answers): the SVCB record, see below.
 
 - `ttl` (`int`; nullable): the TTL if known, otherwise `null`.
 
@@ -239,7 +239,7 @@ The svcb object follows the SVCB specifation in RFC9460.
   
 - `target_name`: (`string`) The TargetName as it appears in the record.
 
-- `params`: The SvcParams, as an object with string values.
+- `params`: (`map[string]string`) The SvcParams, as an map from string to string.
 
 ## Example
 
--- a/data-formats/df-002-dnst.md
+++ b/data-formats/df-002-dnst.md
@@ -239,7 +239,7 @@ The svcb object follows the SVCB specifation in RFC9460.
   
 - `target_name`: (`string`) The TargetName as it appears in the record.
 
-- `params`: (`map[string]string`) The SvcParams, as an map from string to string.
+- `params`: (`map[string]string`) The SvcParams, as a map from string to string.
 
 ## Example
 
--- a/data-formats/df-002-dnst.md
+++ b/data-formats/df-002-dnst.md
@@ -159,13 +159,14 @@ the name because we changed our way of representing getaddrinfo-like resolutions
 
 ```JavaScript
 {
-    // fields currently used as 2022-09-08
+    // fields currently used as 2025-01-31
     "answer_type": "AAAA",
     "asn": 13335
     "as_org_name": "Cloudflare, Inc."
     "hostname": "",
     "ipv4": "1.1.1.1",
     "ipv6": "",
+    "svcb": null,
     "ttl": null,
 
     // specified but unused or deprecated fields
@@ -180,9 +181,9 @@ the name because we changed our way of representing getaddrinfo-like resolutions
 
 - `answer_type`: (`string`) like `Query.query_type`.
 
-- `asn`: (`int`) the Authonomous System Number of the returned ipv4 or ipv6 address.
+- `asn`: (`int`) the Autonomous System Number of the returned ipv4 or ipv6 address.
 
-- `as_org_name`: (`int`) the Authonomous System organisation name of the returned ipv4 or ipv6 address.
+- `as_org_name`: (`string`) the Autonomous System organisation name of the returned ipv4 or ipv6 address.
 
 - `expiration_limit` (`string`; only for SOA answers): the time
 after which this zone should no longer be authoritative.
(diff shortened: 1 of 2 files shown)


nettests/ts-041-ddr.md
```diff
@@ -0,0 +1,138 @@
...
+```JSON
+{
+	"ddr_responses": [],
+    "supports_ddr": false,
+    "resolver": "1.1.1.1",
+    "failure": null
+}
+```
+
+The `resolver` key includes the IP address or domain of the resolver used. In DDR the IP address of the resolver is important information, since the DoE resolvers need to have that IP address in their TLS certificate subjectAltName property to pass the verification step (as per the DDR specification, if the verification fails the resolver should not be used automatically).
```
hellais, line 46: For these keys I believe we should be able to re-use the existing `dnst` data format with some additional extensions.

This is something that we will need also for implementing a full [ECH test](https://github.com/ooni/probe/issues/2840), since that also relies on the lookup of SVCB records.

We could introduce an additional `query_type` and `answer_type` called `SVCB` and encode it as follows:

```
{
    "engine": "getaddrinfo",
    "failure": null,
    "hostname": "_dns.example.net",
    "query_type": "SVCB",
    "raw_response": "dUuBgAABAAEAAAAAB2V4YW1wbGUDY29tAAABAAHADAABAAEAAE2IAARduNgi",
    "resolver_hostname": null,
    "resolver_port": null,
    "resolver_address": null,
    "t0": 0.001145,
    "t": 0.06544,
    "transaction_id": 2,
    "answers": [{
        "answer_type": "SVCB",
        "svcb_raw": "B64ENCODING_OF_SVCB_RECORD",
        "ttl": null
    }],
},
```

If we care also parse the svcb records, we can probably place the `keys` you have now in the `ddr_responses` list inside of the `answers_list` as a list key that contains the parsed values of the SVCB records, but ideally it would be general enough to also support other SVCB records (such as the ECH configuration ones as well).

As for resolver configuration, this is something already supported by the existing data format. During the initial test bootstrap phase we lookup the resolver that probe is using and log it to the `resolver_ip` field (see: https://github.com/ooni/spec/blob/master/data-formats/df-000-base.md). This will match the public IP of the recursive resolver used when using a `getaddrinfo` engine.

In the event of custom resolver, we should use the existing `resolver_address` field to encode it.

frcroth: Thank you for your response. I understand your reasoning and agree with you that all this information should be placed in the preexisting DNS answer type.

Do you think it would be better to have the svcb raw response in the answer or parse the SVCB response completely?
I thought something like this for the probe implementation would work, which is protocol agnostic and does not care about DDR at all.
```
// ArchivalDNSAnswer is a DNS answer.
type ArchivalDNSAnswer struct {
	ASN        int64     `json:"asn,omitempty"`
	ASOrgName  string    `json:"as_org_name,omitempty"`
	AnswerType string    `json:"answer_type"`
	Hostname   string    `json:"hostname,omitempty"`
	IPv4       string    `json:"ipv4,omitempty"`
	IPv6       string    `json:"ipv6,omitempty"`
	TTL        *uint32   `json:"ttl"`
	SVCB       *SVCBData `json:"svcb,omitempty"` // SVCB-specific data
}

// SVCBData represents details of an SVCB record.
type SVCBData struct {
	Priority   uint16            `json:"priority"`
	TargetName string            `json:"target_name"`
	Params     map[string]string `json:"params,omitempty"` // SvcParams key-value pairs
}
```

frcroth: @hellais What do you think about this?

nettests/ts-041-ddr.md
```diff
@@ -0,0 +1,158 @@
...
+        "dohpath":"/dns-query{?dns}",
+        "ipv4hint":"1.1.1.1,1.0.0.1",
+        "ipv6hint":"2606:4700:4700::1111,2606:4700:4700::1001",
+        "port":"443"
+    }
+}
+```
+
+The priority and target_name keys are directly from the SVCB record, where priority specifies the priority among the records at which use this record and target specifies the designated resolver. The key `params` is a map from string to string, that includes the keys that were specified for the SVCB record.
+In the case of this example, the `parmas` map includes this information:
```
hellais, line 66:
```suggestion
In the case of this example, the `params` map includes this information:
```

data-formats/df-002-dnst.md
```diff
@@ -211,8 +212,35 @@ that should be elapsed before the zone should be retried in case of failure.
...
+    }
+}
+```
+The svcb object follows the SVCB specifation in RFC9460.
+
+- `priority`: (`string`) The SvcPriority
+  
+- `target_name`: (`string`) The TargetName as it appears in the record.
+
+- `params`: The SvcParams, as an object with string values.
```
hellais, line 242:
```suggestion
- `params`: (`map[string]string`) The SvcParams, as an map from string to string.
```

hellais (changes requested):
I left some minor edits to the spec, but otherwise look good. Thanks for putting this together.

hellais (approved):
This looks good to me. We should merge it once we land the experiment.

data-formats/df-002-dnst.md
```diff
@@ -211,8 +212,35 @@ that should be elapsed before the zone should be retried in case of failure.
...
+    }
+}
+```
+The svcb object follows the SVCB specifation in RFC9460.
+
+- `priority`: (`string`) The SvcPriority
+  
+- `target_name`: (`string`) The TargetName as it appears in the record.
+
+- `params`: (`map[string]string`) The SvcParams, as an map from string to string.
```
anadahz, line 242:
```suggestion
- `params`: (`map[string]string`) The SvcParams, as a map from string to string.
```

nettests/ts-028-dnscheck.md
```diff
@@ -64,6 +64,19 @@ By manually supplying valid addresses for the domain, users can ensure
...
+out whether it supports [RFC 9462](https://www.rfc-editor.org/rfc/rfc9462)
+Discovery of Designated Resolvers (DDR). The name we query depends on the
+scheme of the input URL:
+
+- for `udp://` and `tcp://` URLs, we query `_dns.resolver.arpa.`, which asks
+the resolver to tell us about the encrypted resolvers it designates;
+
+- for `https://` and `dot://` URLs, we query `_dns.<domain>`, where `<domain>`
+is the hostname in the input URL, which asks the resolver about the
+encrypted DNS services offered by that name.
```
hellais: After discussing this with @DecFox we agreed that this is probably redundant and we should skip doing this, since we are speaking to an encrypted DNS resolver already.

We might want to add this as a richer input configuration option, potentially for SVCB in general.

PR #301, opened 2025-02-18
Oonirun v2.1

hellais:
WIP specification update for OONI Run v2.1

implements https://github.com/ooni/backend/issues/930

### DIFF
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -125,11 +125,23 @@ An OONI Run link descriptor is a JSON file with the following semantics:
   // `array` provides a JSON array of tests to be run.
   "nettests":[{
 
-    // (optional) `array` provides a JSON array of tests to be run.
+    // (optional) `array` provides a JSON array of inputs for the specified test.
     "inputs": [
       "https://example.com/",
       "https://ooni.org/"
     ],
+
+    // (optional) `array` provides a richer JSON array containing targets for the specified experiment.
+    // One of either `inputs` or `targets` can be specified, but not both.
+    "targets": [
+      "input": "https://example.com/",
+      "extra": {
+         "category_code": "HUMR",
+      }
+    ],
+
+   // (optional) string used to specify during creation what is the label for targets to use.
+   "targets_name": "websites_list_prioritized",
 
     // (optional) `map` options arguments provided to the specified test_name
     //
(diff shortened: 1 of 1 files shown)


backends/bk-005-ooni-run-v2.md
```diff
@@ -541,7 +555,99 @@ following JSON body:
...
+
+When the specified OONI Run link contains dynamic targets, it will return the
+generated `targets` list and hence every request might return a different target
+list.
+
+### Request
+
+To retrieve an OONI Run link descriptor, the client issues a request compliant with:
+
+`GET /api/v2/oonirun/links/{oonirun_link_id}/engine-descriptor/{revision}`
```
hellais, line 603: Which of `engine-descriptor` and `full-descriptor` are we currently using in the engine and/or mobile app and in which cases?

cc @aanorbel @DecFox

backends/bk-005-ooni-run-v2.md
```diff
@@ -541,7 +555,105 @@ following JSON body:
...
+### Response status code
+
+Same as 4.3 GET the OONI Run descriptor
+
+### Response body
+
+Same as 4.3 GET the OONI Run descriptor
+
+Note: for dynamically generated run links, this view will only return the
+`target_name` and not the `targets` list.
```
sdsantos, line 583: This way, the users never see the full list of URLs to be tested before installing a Run Link right? Maybe on the mobile clients we need to add a button to see the full list on the web...

backends/bk-005-ooni-run-v2.md
```diff
@@ -125,12 +125,24 @@ An OONI Run link descriptor is a JSON file with the following semantics:
...
+    // One of either `inputs` or `targets` can be specified, but not both.
+    "targets": [
+      "input": "https://example.com/",
+      "extra": {
+         "category_code": "HUMR",
+      }
+    ],
+
+   // (optional) string used to specify during creation what is the label for targets to use.
+   "targets_name": "websites_list_prioritized",
```
sdsantos, line 144: This will come from a fixed list of options right?

sdsantos (commented):
This look to be breaking changes to the schema. Should the endpoints become `.../api/v2.1/...`?

backends/bk-005-ooni-run-v2.md
hellais: Yeah, this is by design though, given that for dynamic test lists (which are currently only derived from the test lists), we cannot know the full list ahead of time as it will change every time you run it.

From the perspective of the UI, we can either show a label that explains these are dynamic and refer to the the test-lists repo to view them. Alternatively we can fetch a batch of URLs from the API endpoint and display that, but make it clear it's just a sample and might change.

hellais:
>This look to be breaking changes to the schema. Should the endpoints become .../api/v2.1/...?

The concern here is that `inputs` might be empty and the clients are not able to read the `targets` and `targets_name`

hellais:
I think I have come up with a better solution to handle older clients in: https://github.com/ooni/spec/pull/301/commits/6f701788b501b2ee47545581fb113c5d999af043.

Basically we don't need to have a new key, but instead we just handle extra metadata for inputs through `inputs_extra`. That way if clients don't understand `inputs_extra`, they will still be able to read `inputs` like normal.

It also should make the implementation easier and not have to handle multiple different cases depending on client support.

backends/bk-005-ooni-run-v2.md
```diff
@@ -551,13 +667,13 @@ Authentication for this endpoint is optional.
 
 To retrieve an OONI Run link descriptor, the client issues a request compliant with:
 
-`GET /api/v2/oonirun_links?only_latest=true&only_mine=true&include_archived=true`
+`GET /api/v2/oonirun/links?only_latest=true&is_mine=true&is_expired=true`
```
hellais, line 670:
```suggestion
`GET /api/v2/oonirun/links?is_mine=true&is_expired=true`
```

LDiazN (approved):
Approved!

PR #310, opened 2026-08-05
oonirun: specify configuration model, target identity and attribution

hellais:
Three additions and a set of corrections to the OONI Run v2 spec, motivated by the changes to support oonirun v2.1 in miniooni, the plan to rebuild the instant-messaging cards as links composed from web_connectivity building blocks, and by the data pipeline's requirements on provenance and series stability.

It also folds into the spec some implicit changes introduce in https://github.com/ooni/probe-cli/pull/1812 and reserves some special keys in `inputs_extra` on which specific semantics are attached.

In theory nothing in this spec should require additional engine or backend changes, except for `ooni_run_link_revision` (should be easy) and `ooni_run_attempt` (might be harder). 

### Below is a more comprehensive description of the changes

Configuration model (3.1). Reinstates the nettest-level `options` object alongside the per-input `inputs_extra` overlay, with defined merge semantics: engine defaults, then options, then inputs_extra[i], shallow merge. Without a nettest-level object, input-less nettests had no configuration surface at all and shared settings had to be duplicated into every inputs_extra entry. Backends validate option names at CREATE so a typo fails loudly instead of silently doing nothing in the field; probes ignore unknown names so old probes keep working. The safe_ prefix convention, which the spec used but never documented, is now stated together with its scrubbing rule. The effective per-input configuration is recorded in the measurement, so analysis can condition on what actually ran.

Target identity (3.2). Inputs are addresses and may rotate freely between revisions; targets are durable names, expressed as service roles and never as addresses, kept in a shared append-only registry. Inputs sharing a target_id form a redundant pool; the breaks_service flag marks a target whose blocking breaks the whole service for the user, the way WhatsApp registration does regardless of the chat pool. One declaration drives the probe-side card status and the pipeline-side aggregation, which is what lets an IM card decompose into web_connectivity runs without collapsing into per-hostname series that break at every infrastructure rename.

Measurement attribution (5.0). Probes annotate every measurement with the link id, the revision and a per-run attempt id, which is what scopes aggregates to a campaign and reconstructs a composed link as one logical check. Consent is defined to bind to revisions rather than resolutions: a volunteer consents to the named list, not to the specific inputs it yields on a given day, so dynamic-list churn no longer re-prompts while every definition change still does. Resolutions themselves are ephemeral by design, since retaining every prioritization output would grow without bound; the measurements submitted under one attempt are the durable record of what was served, and the spec states the accepted tradeoff that the unmeasured remainder is not reconstructible.

Corrections: duplicate 4.4 heading renumbered (engine descriptor is now 4.5, LIST is 4.6, and it is titled POST, which is what it specifies); stale v1 API path and run.ooni.io addresses updated; expired/archived terminology unified; User-Agent field list typo fixed.

PR companion to OONI Runv2.1 changes in miniooni: https://github.com/ooni/probe-cli/pull/1812

### DIFF
--- a/backends/bk-005-ooni-run-v2.md
+++ b/backends/bk-005-ooni-run-v2.md
@@ -1,7 +1,7 @@
 # OONI Run v2 specification
 
 -   author: Norbel Ambanumben, Arturo Filastò
--   version: 2024.02.23
+-   version: 2026.08.05
 -   status: release-candidate
 
 This document provides a functional specification for OONI Run.
@@ -36,19 +36,19 @@ sequenceDiagram
     actor Volunteer
 
     CampaignOrganizer->>OONIAPI: Create OONI Run link to run web_connectivity with http://example.com
-    OONIAPI->>CampaignOrganizer: OONI Run link: https://run.ooni.io/v2/deadbeef
-    CampaignOrganizer-->>Volunteer: Hey, can you open the link https://run.ooni.io/v2/deadbeef with OONI Probe?
+    OONIAPI->>CampaignOrganizer: OONI Run link: https://run.ooni.org/v2/deadbeef
+    CampaignOrganizer-->>Volunteer: Hey, can you open the link https://run.ooni.org/v2/deadbeef with OONI Probe?
     Volunteer-)OONIAPI: What is the descriptor for OONI Run link with ID "deadbeef"
     OONIAPI->>Volunteer: Here is the descriptor for "deadbeef"
 ```
 
 It's important to notice how, in the event that the user has the OONI Probe app
-installed, a web request to `https://run.ooni.org/io/v2/deadbeef` will never be
+installed, a web request to `https://run.ooni.org/v2/deadbeef` will never be
 issued on the network, but rather the metadata encoded in the URL itself is used
 to retrieve the OONI Run descriptor from a different OONI API endpoint.
 
 When a Volunteer taps on an OONI Run link (in the above example
-`https://run.ooni.io/v2/deadbeef`) the OONI Probe app is opened and a they are
+`https://run.ooni.org/v2/deadbeef`) the OONI Probe app is opened and a they are
 presented with the metadata of the OONI Run link as well as the nettests it is
 configured with.
 At this point, assuming they feel confortable with running the nettests they see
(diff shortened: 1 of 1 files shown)


backends/bk-005-ooni-run-v2.md
```diff
@@ -724,7 +839,26 @@ following JSON body:
...
+# 5.0 Measurement attribution
+
+Probes MUST annotate every measurement produced while running an OONI Run link
+with the following annotations:
+
+| annotation | value |
+| --- | --- |
+| `ooni_run_link_id` | the OONI Run link id |
+| `ooni_run_link_revision` | the revision of the descriptor that was run |
+| `ooni_run_attempt` | a random UUID minted once per link run and shared by all measurements produced by that run |
```
hellais, line 922: If this is too complex to add now, we can defer it to future work

backends/bk-005-ooni-run-v2.md
```diff
@@ -180,6 +222,140 @@ card. The OONI Run descriptor, as specified from the link creator, is saying
...
+  corresponding input.
+
+The effective configuration for input *i* is computed by a shallow, field-wise
+merge: engine defaults, then `options`, then `inputs_extra[i]`, with later
+values winning. Nested objects are replaced, not merged.
+
+Option names and their meaning are defined by each nettest. The following
+rules keep authoring mistakes loud and probes forward-compatible:
+
+* Backends MUST reject at CREATE and UPDATE time (with a `4xx`) any option
```
hellais:
```suggestion
* Backends MAY reject at CREATE and UPDATE time (with a `4xx`) any option
```

backends/bk-005-ooni-run-v2.md
```diff
@@ -120,25 +135,52 @@ An OONI Run link descriptor is a JSON file with the following semantics:
...
+        "category_code": "HUMR",
     }],
 
-   // (optional) string used to specify during creation that the input list should be dynamically generated. The semantics of each string is up to the backend implementation.
-   "targets_name": "websites_list_prioritized",
+   // (optional) `string` naming a backend-generated dynamic input list (see
+   // section 3.3). Names are owned and validated by the backend; probes MUST
+   // treat them as opaque and MUST NOT validate them against any hardcoded
+   // list, so that new names can be introduced without a client update.
+   "targets_name": "citizenlab/test_lists",
```
hellais:
```suggestion
   "targets_name": "websites_list_prioritized",
```

backends/bk-005-ooni-run-v2.md
```diff
@@ -180,6 +222,140 @@ card. The OONI Run descriptor, as specified from the link creator, is saying
...
+outcome of a composed link, for example rendering a single card status for a
+messaging app whose link measures several pools and services.
+
+## 3.3 Dynamic target lists
+
+`targets_name` names an input list that the backend generates at resolution
+time (see 4.5) instead of the list being written into the descriptor. Two
+namespaces are defined today:
+
+* `citizenlab/test_lists`: the community test lists, prioritized and sorted
```
hellais:
```suggestion
* `websites_list_prioritized`: the community test lists, prioritized and sorted
```

hellais:
We should update the spec to align to what the backend currently does:
* Change the `target_names` label to websites_list_prioritized
* Remove the MUST validate related to `inputs_extra` target_ids
* Remove the validation of `breaks_service`
* Remove references to target_registry, until we have it ready
* Remove validation of `options` inside the backend, unless we are able to programmatically pull them from the go engine

